PR #65, opened 2019-08-20
Handle timezone from TZ variable

Turgon37:
Timezone should not be set in hard, but can be overrided by common TZ environment variable

### DIFF
--- a/ara/server/settings.py
+++ b/ara/server/settings.py
@@ -186,7 +186,7 @@ AUTH_PASSWORD_VALIDATORS = [
 ]
 
 USE_TZ = True
-TIME_ZONE = "UTC"
+TIME_ZONE = os.getenv('TZ', 'UTC')
 
 # We do not currently support internationalization and localization, turn these
 # off.


dmsimard:
Hi @Turgon37 and thanks for looking at this !

You're right, we shouldn't be hardcoding the timezone.

We're managing Django settings with [dynaconf](https://github.com/rochacbruno/dynaconf/blob/master/docs/guides/django.md) which takes care of exposing the settings over environment variables and in configuration files.

So, actually, what we'd want to do here is:
```python
TIME_ZONE = settings.get("TIME_ZONE", "UTC")
```
And then later expose that value in the default settings.yaml by adding a key in this dict: https://github.com/ansible-community/ara/blob/f252841e7ba9ea6918e231dfee178ae7384e0f6f/ara/server/settings.py#L240-L262

We unfortunately can't quite accept PRs over GitHub yet because the project code review is set up in Gerrit.
We do have an [issue](https://github.com/ansible-community/ara-infra/issues/4) to set up automation to allow PRs but in the meantime, would you like to submit this to Gerrit ?

We have documentation for contributors [here](https://ara.readthedocs.io/en/latest/contributing.html) and I can help getting you started. Feel free to ping me on [Slack or IRC](https://ara.recordsansible.org/community/) if you'd like to chat !

Turgon37:
closed in favor of https://review.opendev.org/#/c/678348/1/ara/server/settings.py

PR #66, opened 2019-08-20
Add placeholder to prepare ordering handling

Turgon37:
For now the api result the playbooks ordered by the postgresql defaut order (primary key). So the first records are the most older ones.
I want to propose an alternate ordering in my docker image https://github.com/Turgon37/docker-ansible-araserver/pull/1 so this PR will allow people to use "sed" to change ordering in code. I know this is a bad practice, but I do not see the best way to do this.

### DIFF
--- a/ara/api/models.py
+++ b/ara/api/models.py
@@ -39,6 +39,7 @@ class Duration(Base):
 
     class Meta:
         abstract = True
+        # PLACEHOLDER_DURATION_ORDERING
 
     started = models.DateTimeField(default=timezone.now)
     ended = models.DateTimeField(blank=True, null=True)
@@ -72,7 +73,7 @@ class Playbook(Duration):
     is tied back to this one playbook.
     """
 
-    class Meta:
+    class Meta(Duration.Meta):
         db_table = "playbooks"
 
     # A playbook in ARA can be running (in progress), completed (succeeded) or failed.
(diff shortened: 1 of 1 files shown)


Turgon37:
According to documentation https://docs.djangoproject.com/en/2.2/topics/db/models/#abstract-base-classes the abstract attribute is not herited in children classes

dmsimard:
Hi @Turgon37 and thanks for your time !

We do need to set up ordering to be customizable but I don't think we need to do it through the models.
The API is provided through django-rest-framework and they have documentation on how to set up ordering here: https://www.django-rest-framework.org/api-guide/filtering/#orderingfilter

If you would like to look into this, you'll find the API viewsets here: https://github.com/ansible-community/ara/blob/master/ara/api/views.py

Turgon37:
Thanks for the reply, I will wait for this feature to be implemented.
But I'm worried because of two things,
1, I've not seen any database index on column other than primary keys, so I can suppose the sort will be expensive to the RDBMS on ara installations with amount of records.
2. I've performed some test with Django viewset and I'm notice that the sort is performed by the code not by the database. In fact, by using the model's "ordering" attribute the sort is performed with the ORDER BY SQL clause.

Please can you consider theses into your development ?

I'm starting using ARA, and I found it very usefull because Ansible lacks of theses feature until now :)

Turgon37:
I've published this review https://review.opendev.org/#/c/678347/1/ara/api/models.py

dmsimard:
@Turgon37 hi!

I am closing this PR because we just merged a change that implements sorting with https://github.com/ansible-community/ara/commit/b02d8279a4101bf1bf30eaee584da07209687435

The change will land in ara 1.3.

PR #67, opened 2019-08-24
Removed unneeded autoescape.

apollo13:
Django does autoescaping by default. I also verified by putting `<script>alert(document.cookie);</script>` in a playbook argument and it ends up escaped as expected. If your smoke test where able to find something here I'd be very much interested to know what :)


dmsimard:
@apollo13 Can you send this to gerrit ? We have an integration test that exercises this so we can tell.

We aren't set up to accept PRs yet.

apollo13:
Ah :) https://review.opendev.org/#/c/680644/

PR #70, opened 2019-09-19
Ensure endpoint doesn't contain trailing slash

sshnaidm:
If api_server is configured with trailing slash as "http://localhost/" it causes failures and it's difficult to investigate. Ensure it doesn't happen and endpoint is always without trailing slash.


sshnaidm:
Done via gerrit: https://review.opendev.org/683130

PR #78, opened 2019-10-13
Remove leading newline from default settings template

berendt:


dmsimard:
@berendt hi and thanks for the PR! Unfortunately we aren't set up to accept pull requests in GitHub yet. Could you submit this through [gerrit](https://ara.readthedocs.io/en/latest/contributing.html#issues-and-pull-requests) instead ?

berendt:
Yes, of course. Makes it easier for me, too. I just thought after moving the repository to ansible-community we now take PRs.

dmsimard:
@berendt awesome, thank you very much !

PR #97, opened 2019-12-03
Fixed typo in ara.setup.env

sherve-korem:
This typo breaks local modules (in a playbook's ./library for ex.) by making modules mistaken for action plugins in Ansible 2.9.1.

### DIFF
--- a/ara/setup/env.py
+++ b/ara/setup/env.py
@@ -24,7 +24,7 @@ from . import action_plugins, callback_plugins
 
 exports = """
 export ANSIBLE_CALLBACK_PLUGINS=${{ANSIBLE_CALLBACK_PLUGINS:-}}${{ANSIBLE_CALLBACK_PLUGINS+:}}{}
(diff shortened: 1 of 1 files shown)


dmsimard:
Hi @sherve-korem and thanks for finding this !
Unfortunately we aren't set up to accept pull requests on GitHub yet because this repository is a mirror.

We have [documentation](https://ara.readthedocs.io/en/latest/contributing.html) on how to send a patch and while I would appreciate if you could submit your patch to gerrit, I understand that this may be a few hoops to jump through just to fix a typo.

Let me know if you would like to do it or if you would prefer I submit the patch on your behalf.

Thanks !

sherve-korem:
Hi @dmsimard , I would prefer that you submit the patch indeed, as I don't have any of the required setup on hand.

dmsimard:
Thanks @sherve-korem ! I've submitted the patch as https://review.opendev.org/#/c/697172/

Unfortunately, this won't be in the 1.3 release from today but it will be in the next dot release.
It will be available from git master as soon as the patch lands.

PR #114, opened 2020-02-13
ui: use relative links for pagination

flowerysong:
These were being generated as e.g. `http://localhost:8000/?limit=100&offset=100`, while the rest of the links in the UI are relative paths like `playbook/1.html`.

Since I'm not accessing the UI from localhost this caused some breakage.

### DIFF
--- a/ara/ui/pagination.py
+++ b/ara/ui/pagination.py
@@ -26,18 +26,42 @@ class LimitOffsetPaginationWithLinks(LimitOffsetPagination):
     """
     Extends LimitOffsetPagination to provide links
     to first and last pages as well as the limit and offset, if available.
+    Generates relative links instead of absolute URIs.
     """
+
+    def get_next_link(self):
+        if self.offset + self.limit >= self.count:
+            return None
+
+        url = self.request.get_full_path()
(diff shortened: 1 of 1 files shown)


dmsimard:
Hi @flowerysong and thanks for the patch !

Unfortunately, we aren't set up for accepting pull requests ([yet](https://github.com/ansible-community/ara-infra/issues/4)) because this repository is a mirror and code review is done in [Gerrit](https://review.opendev.org/#/q/project:recordsansible/ara).

If you'd like to send the patch through Gerrit, there's a few hoops to jump through that are in the docs here: https://ara.readthedocs.io/en/latest/contributing.html

I realize this can be cumbersome and I can submit the commit on your behalf if you'd like.

Thanks again !

flowerysong:
@dmsimard It would be useful to put that in the README (it currently points to GitHub for bugs, issues, and enhancements) and also in a PR template.

dmsimard:
@flowerysong agreed, good idea. Thank you for taking the time to send the patch, much appreciated.
I'm taking some much needed time off right now but I'll merge it asap.

PR #206, opened 2021-02-25
ui: First iteration of the new host index

dmsimard:
### Exported from gerrit: https://review.opendev.org/c/recordsansible/ara/+/682778

Change-Id: Ib72729c2ead01a48ae8639007161b98c99a80a2

This is a work in progress that adds a host index a bit like the playbook index so you can browse, search and query for hosts instead of playbooks.

### DIFF
--- a/ara/ui/views.py
+++ b/ara/ui/views.py
@@ -53,6 +53,48 @@ class Index(generics.ListAPIView):
         # fmt: on
 
 
+class HostIndex(generics.RetrieveAPIView):
+    """
+    Returns a list of latest record for each host
+    """
+
+    queryset = models.DistinctHost.objects.all()
+    filterset_class = filters.HostFilter
+    renderer_classes = [TemplateHTMLRenderer]
+    pagination_class = LimitOffsetPaginationWithLinks
+    template_name = "host_index.html"
+
+    def get(self, request, *args, **kwargs):
+        search_form = forms.HostSearchForm(request.GET)
+
+        query = self.filter_queryset(self.queryset.all().order_by("-name"))
+        page = self.paginate_queryset(query)
+        if page is not None:
+            serializer = serializers.ListDistinctHostSerializer(page, many=True)
+        else:
+            serializer = serializers.ListDistinctHostSerializer(query, many=True)
+        response = self.get_paginated_response(serializer.data)
+
+        if self.paginator.count > (self.paginator.offset + self.paginator.limit):
+            max_current = self.paginator.offset + self.paginator.limit
+        else:
+            max_current = self.paginator.count
+        current_page_results = "%s-%s" % (self.paginator.offset + 1, max_current)
+
+        # We need to expand the search card if there is a search query, not considering pagination args
+        search_args = [arg for arg in request.GET.keys() if arg not in ["limit", "offset"]]
+        expand_search = True if search_args else False
+
+        return Response(dict(
+            current_page_results=current_page_results,
+            data=response.data,
+            expand_search=expand_search,
+            page="host_index",
+            search_form=search_form,
+            static_generation=False
+        ))
+
+
 class Playbook(generics.RetrieveAPIView):
     """
     Returns a page for a detailed view of a playbook
--- a/ara/ui/views.py
+++ b/ara/ui/views.py
@@ -67,7 +67,7 @@ class HostIndex(generics.RetrieveAPIView):
     def get(self, request, *args, **kwargs):
         search_form = forms.HostSearchForm(request.GET)
 
-        query = self.filter_queryset(self.queryset.all().order_by("-name"))
+        query = self.filter_queryset(self.queryset.all().order_by("name"))
         page = self.paginate_queryset(query)
         if page is not None:
             serializer = serializers.ListDistinctHostSerializer(page, many=True)
--- a/ara/ui/views.py
+++ b/ara/ui/views.py
@@ -58,8 +58,6 @@ class HostIndex(generics.RetrieveAPIView):
     Returns a list of latest record for each host
     """
 
-    queryset = models.DistinctHost.objects.all()
-    filterset_class = filters.HostFilter
     renderer_classes = [TemplateHTMLRenderer]
     pagination_class = LimitOffsetPaginationWithLinks
     template_name = "host_index.html"
@@ -67,13 +65,27 @@ class HostIndex(generics.RetrieveAPIView):
     def get(self, request, *args, **kwargs):
         search_form = forms.HostSearchForm(request.GET)
 
-        # Sort by updated by default so we have the most recently updated at the top
-        query = self.filter_queryset(self.queryset.all().order_by("-updated"))
-        page = self.paginate_queryset(query)
-        if page is not None:
-            serializer = serializers.DetailedDistinctHostSerializer(page, many=True)
+        # Default is DistinctHost (by not requiring "?latest=true") but accept false to
+        # return all hosts
+        if "latest" in request.GET and request.GET["latest"] == "false":
+            queryset = models.Host.objects.all()
+            serializer_type = "DetailedHostSerializer"
+            filter_type = "HostFilter"
+            # TODO: Is there a cleaner way ? Doing this logic in the template seemed complicated.
+            checkbox_status = "checked"
         else:
-            serializer = serializers.DetailedDistinctHostSerializer(query, many=True)
+            queryset = models.DistinctHost.objects.all()
+            serializer_type = "DetailedDistinctHostSerializer"
+            filter_type = "DistinctHostFilter"
+            checkbox_status = ""
+
+        # Sort by updated by default so we have the most recently updated at the top
+        query = getattr(filters, filter_type)(request.GET, queryset=queryset)
+        page = self.paginate_queryset(query.qs.all().order_by("-updated"))
+        if page is not None:
+            serializer = getattr(serializers, serializer_type)(page, many=True)
+        else:
+            serializer = getattr(serializers, serializer_type)(query, many=True)
         response = self.get_paginated_response(serializer.data)
 
         if self.paginator.count > (self.paginator.offset + self.paginator.limit):
@@ -88,6 +100,7 @@ class HostIndex(generics.RetrieveAPIView):
 
         # fmt: off
         return Response(dict(
+                checkbox_status=checkbox_status,
                 current_page_results=current_page_results,
                 data=response.data,
                 expand_search=expand_search,
--- a/ara/ui/views.py
+++ b/ara/ui/views.py
@@ -104,13 +104,13 @@ class HostIndex(generics.RetrieveAPIView):
 
         # fmt: off
         return Response(dict(
-                checkbox_status=checkbox_status,
-                current_page_results=current_page_results,
-                data=response.data,
-                expand_search=expand_search,
-                page="host_index",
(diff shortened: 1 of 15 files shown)


kfiresmith:
Killer feature IMO.  The more ARA can do to provide CMDB-like features the better IMO, especially with Ansible-CMDB largely stale as a project.

dmsimard:
Hey @kfiresmith and thanks for the feedback o/

It's also a feature I look forward to but unfortunately the ongoing UI refactor has been taking more time and effort than anticipated so this PR won't make the cut for the upcoming 1.5.6 release. I'll have time for it once 1.5.6 is out which should be soon enough.

dmsimard:
BTW: This is still a work in progress and I haven't pushed an update to the PR yet because it depends on #247 to be merged first. I have a WIP branch of this change including the contents of 247 here: https://github.com/dmsimard/ara/tree/wip_host_index

dmsimard:
I've updated the PR and rebased it on top of master now that #247 has been merged. It still needs work so I'll iterate a bit on it in the near future but feel free to review/comment in the meantime. The current state of the PR (as of this comment) is deployed on a demo instance: https://dev.demo.recordsansible.org/hosts

dmsimard:
Some TODOs/questions/thoughts:
- I initially set the default sort order to alphabetical by name but I think we should set the default sort order to be by date with the most recent report at the top in order to be consistent with the playbook reports page
- Need to add a sorting icon (and functionality) to the name column
- The "last 60 minutes", "last 24hrs", etc buttons in the filtering box do not work yet (copy/pasted placeholder from playbook implementation)
- I've cherry-picked a list of host facts that seem worthwhile to display if they are available: https://github.com/ansible-community/ara/pull/206/files#diff-10d83b6fceb43d7b51fda0f67e2270a85d9a95e6ca3a33a17b869a70ac49d083. Searching capability may come later. Does this seem like a good start ? Any other facts that we might want to add ?
- Is the column order right (report > name > last result > host facts) ? It bothers me a bit but I'm not sure what's the best order either right now.

hille721:
Here are my two cents:

> - I initially set the default sort order to alphabetical by name but I think we should set the default sort order to be by date with the most recent report at the top in order to be consistent with the playbook reports page

I really like the current sort order (alphabetical by name) because on a host view it is what I expect.

> * I've cherry-picked a list of host facts that seem worthwhile to display if they are available: https://github.com/ansible-community/ara/pull/206/files#diff-10d83b6fceb43d7b51fda0f67e2270a85d9a95e6ca3a33a17b869a70ac49d083. Searching capability may come later. Does this seem like a good start ? Any other facts that we might want to add ?

I don't have currently any more fact to add. Thus fine for me.

> * Is the column order right (report > name > last result > host facts) ? It bothers me a bit but I'm not sure what's the best order either right now.

Honestly I haven't thought about that, but I like it as it is currently. 

But what I really want to see here:
* If I click on the report button, I will get the last result, thus this is already done
* But If I click on the hostname, I want to see a a list with all reports of this host (thus API speaking: https://dev.demo.recordsansible.org/api/v1/hosts?name=aio1_cinder_api_container-db31a68f)
That would be AWESOME!

dmsimard:
> > * I initially set the default sort order to alphabetical by name but I think we should set the default sort order to be by date with the most recent report at the top in order to be consistent with the playbook reports page
> 
> I really like the current sort order (alphabetical by name) because on a host view it is what I expect.

I thought about that and I like that too, but it also means that it will eventually always be the same hosts that show up at the top due to their name.
Pretend you want to look at a host that starts with the letter w (webserver), you could end up /always/ needing to go to another page or search for it.

I definitely want to keep the sorting by name available but I was wondering whether sorting by last playbook run date had more value. On the first page you would have the hosts that have most recently "checked in" and I think that's pretty relevant.

> > * I've cherry-picked a list of host facts that seem worthwhile to display if they are available: https://github.com/ansible-community/ara/pull/206/files#diff-10d83b6fceb43d7b51fda0f67e2270a85d9a95e6ca3a33a17b869a70ac49d083. Searching capability may come later. Does this seem like a good start ? Any other facts that we might want to add ?
> 
> I don't have currently any more fact to add. Thus fine for me.

:+1:

> > * Is the column order right (report > name > last result > host facts) ? It bothers me a bit but I'm not sure what's the best order either right now.
> 
> Honestly I haven't thought about that, but I like it as it is currently.

Yeah, I think I'm just not used to it -- it kind of grew on me now :p

> But what I really want to see here:
> 
>     * If I click on the report button, I will get the last result, thus this is already done

:tada: 

>     * But If I click on the hostname, I want to see a a list with all reports of this host (thus API speaking: https://dev.demo.recordsansible.org/api/v1/hosts?name=aio1_cinder_api_container-db31a68f)

I did experiment trying to do something like that but I didn't like how I had implemented it: https://github.com/ansible-community/ara/pull/247#issuecomment-864452266

I'll think about it but I'm open to suggestions :p

>       That would be AWESOME!

It would.

dmsimard:
> > > * I initially set the default sort order to alphabetical by name but I think we should set the default sort order to be by date with the most recent report at the top in order to be consistent with the playbook reports page
> > 
> > 
> > I really like the current sort order (alphabetical by name) because on a host view it is what I expect.
> 
> I thought about that and I like that too, but it also means that it will eventually always be the same hosts that show up at the top due to their name.
> Pretend you want to look at a host that starts with the letter w (webserver), you could end up /always/ needing to go to another page or search for it.
> 
> I definitely want to keep the sorting by name available but I was wondering whether sorting by last playbook run date by default had more value. On the first page you would have the hosts that have most recently "checked in" and I think that's pretty relevant.
> 

It also means that if there are any hosts on which playbooks are running currently, they will be shown at the top (ordering by updated).

dmsimard:
Needs more work but I did a quick iteration to add columns for the playbook status and playbook duration, I think they're helpful context to have and it turns out that we pretty much have the same four columns for the playbook list and host list now:

Playbook list page:

Selection_171

Host list page:

Selection_172

dmsimard:
So we don't forget, it would also be nice to be able to filter hosts by a fact's value like suggested here: https://github.com/ansible-community/ara/pull/247#issuecomment-864582635

For example, there could be two search fields: ``fact`` and ``value`` so you could search for ``ansible_distribution`` as fact and then ``Fedora`` to only return hosts that have a match.

If we make the host fact labels search when you click on them, there's gotchas:
-  "Fedora 34" is actually a combination of ``ansible_distribution`` and ``ansible_distribution_version`` (can we search for N fact=ansible_something&factvalue=value ?)
- The uptime search would make little sense (although, sort by uptime maybe sometime ?)

dmsimard:
I got a toggle between all hosts and distincthosts that I'm happier with.

I'll clean it up and push it later but it looks like this:
Peek 2021-07-24 16-43

dmsimard:
Pushed a commit to add the latest toggle, search by status and fixed date ranges buttons. Looks like this:
Peek 2021-07-24 20-39

dmsimard:
Fixed sorting by date and added sorting by name:
Peek 2021-07-26 11-38

dmsimard:
I've gone through the TODOs I had listed in a previous comment: https://github.com/ansible-community/ara/pull/206#issuecomment-880130746 and in addition I've included the toggle for "latest only/all hosts".

This is "good enough" for a first iteration and we can improve it in future PRs/patches/releases.
There may be adjustments, fixes or improvements before 1.5.7 goes out but this is the bulk of it.

Support for searching facts and fact values (mentioned [here](https://github.com/ansible-community/ara/pull/206#issuecomment-885992773)) did not make the cut. It's a nice to have feature for sure but not a blocker at all so I've created an issue in order to take care of it later: https://github.com/ansible-community/ara/issues/297

ara/ui/views.py
```diff
@@ -53,6 +53,68 @@ def get(self, request, *args, **kwargs):
...
+            filter_type = "DistinctHostFilter"
+            checkbox_status = ""
+
+        # Sort by updated by default so we have the most recently updated at the top
+        order = "-updated"
+        if "order" in request.GET:
+            order = request.GET["order"]
+
+        query = getattr(filters, filter_type)(request.GET, queryset=queryset)
+        page = self.paginate_queryset(query.qs.all().order_by(order))
```
dmsimard, line 88: I think there might be a bug here.

It is ordering by ``-updated`` which is the ``updated`` field of the latesthost, *NOT* of the actual host.
This would be fine off of a fresh database since the "updated" field would continuously get updated.

However, because we are not supplying the "created" and "updated" timestamps when running the SQL migration, they are set to "now" instead of "whenever it was for that host" so sorting by date on a database that has run migrations leads to results that are not completely wrong but a bit weird.

dmsimard: Visualized here after running ``ara-manage migrate``:
Screenshot from 2021-07-26 18-21-47

I tried to force the ``created`` and ``updated`` fields to match the child host inside the sql migration and the best I've got was the created field because, well, doing an update updates the updated field to "now" :P

Screenshot from 2021-07-26 18-28-45

This is tricky. I'll think about it.

dmsimard: Additional note: this is only an issue for the view with the latest playbook per host because the updated field is not accurate. When enabling the toggle to view all hosts, the data is ordered correctly.

dmsimard: This makes me wonder: should we maybe remove the ``auto_now`` and ``auto_now_add`` on the model ?
https://github.com/ansible-community/ara/blob/7c5988ae7d0c1c1779b2437ff7c52e04e265c52d/ara/api/models.py#L256-L258

For example, if a host gets deleted, the latesthost reference will be updated and thus bumping the updated field but it doesn't mean the host that has taken it's place has been updated recently.

Sorting by created doesn't exactly work either because the latesthost will probably always have the same created date (unless it gets deleted and re-created).

If we remove the auto_now and auto_now_add, we would be "free" to put the dates we want in the sql migration as well as in the creation/deletion management:
https://github.com/ansible-community/ara/blob/7c5988ae7d0c1c1779b2437ff7c52e04e265c52d/ara/api/views.py#L108-L113

https://github.com/ansible-community/ara/blob/7c5988ae7d0c1c1779b2437ff7c52e04e265c52d/ara/api/serializers.py#L341

hille721: in my opinion the updated and created fields of the DistinctHost/ LatestHost objects are describing when this object was firstly created and when it was updated. It should NOT equals to the updated and created field of the related host object.

hille721: > Additional note: this is only an issue for the view with the latest playbook per host because the updated field is not accurate. When enabling the toggle to view all hosts, the data is ordered correctly.

I think the cleanest solution is if you are using the updated field of the related host object in the view and not the updated field of the DistinctHost/LatestHost object itself.
Probably more difficult, but a clean solution ;)

dmsimard: @hille721 fair enough -- if the intent is to sort by the updated field of the child hosts, is there a way to do it without relying on the updated field of the latesthost ?

hille721: this should work:

```
.order_by('latest__updated')
```

dmsimard: I didn't know you could sort by nested fields like that, I will try it !

dmsimard: It works !

I've sent a PR to fix it: https://github.com/ansible-community/ara/pull/300. Thank you :)

PR #208, opened 2021-02-26
zuul: add integration test jobs back in

dmsimard:
These are the same jobs as before, just adding them back in.


dmsimard:
recheck

PR #209, opened 2021-02-26
output context with newlines in plaintext diff

aorfanos:
The problem with current implementation is that diff output in e.g. bash scripts is showing up as a big string instead of formatted lines (I'm using ARA 1.4.3). Using the `diff_format_string` function should output a formatted result.

### DIFF
--- a/ara/ui/templatetags/diff_result.py
+++ b/ara/ui/templatetags/diff_result.py
@@ -13,7 +13,7 @@ def diff_format_string(plaintext):
     output = []
     for line in plaintext.splitlines():
         output.append(line)
-    return '\n'.join(output)
+    return "\n".join(output)
 
 
 def render_diff(before="", after="", before_header="before", after_header="after"):
--- a/ara/ui/templatetags/diff_result.py
+++ b/ara/ui/templatetags/diff_result.py
@@ -8,6 +8,11 @@ from django import template
 
 register = template.Library()
 
+def diff_format_string(plaintext):
+    output = []
+    for line in plaintext.splitlines():
+        output.append(line)
+    return '\n'.join(output)
 
 def render_diff(before="", after="", before_header="before", after_header="after"):
     """
@@ -24,13 +29,12 @@ def render_diff(before="", after="", before_header="before", after_header="after
         )
     else:
         return difflib.unified_diff(
-            before.splitlines(),
-            after.splitlines(),
+            diff_format_string(before),
+            diff_format_string(after),
             fromfile=before_header,
             tofile=after_header
         )
     # fmt: on
-
 
 @register.filter(name="diff_result")
 def diff_result(diff):
--- a/ara/ui/templatetags/diff_result.py
+++ b/ara/ui/templatetags/diff_result.py
@@ -8,11 +8,13 @@ from django import template
 
 register = template.Library()
 
+
 def diff_format_string(plaintext):
     output = []
     for line in plaintext.splitlines():
         output.append(line)
     return '\n'.join(output)
+
 
 def render_diff(before="", after="", before_header="before", after_header="after"):
     """
@@ -35,6 +37,7 @@ def render_diff(before="", after="", before_header="before", after_header="after
             tofile=after_header
         )
     # fmt: on
+
 
 @register.filter(name="diff_result")
 def diff_result(diff):


ara/ui/templatetags/diff_result.py
```diff
@@ -24,8 +31,8 @@ def render_diff(before="", after="", before_header="before", after_header="after
         )
     else:
         return difflib.unified_diff(
-            before.splitlines(),
-            after.splitlines(),
```
dmsimard, line 28: Thanks for the PR @aorfanos !

Could you provide a reproducer I could test with to see what you mean with the big string issue ?

I was wondering if this could be written as something like ``"\n".join(before.splitlines())`` without needing a function but in any case, it doesn't seem like it has the intended effect on other diffs like this one before:
Screenshot from 2021-02-26 13-29-31


and after:
Screenshot from 2021-02-26 13-31-58

aorfanos: Using the `file` module yields the following diff, which is fixed by the addition in this PR (I have formatted it so it doesn't show in one line):

```
[
    {
        "after": "# Managed by Ansible \n\ndatadir=\"/var/lib/proxysql\"\n\nadmin_variables =
 \n{\n    admin_credentials=\"admin:admin\"\n    mysql_ifaces=\"0.0.0.0:6032\"\n
cluster_username=\"admin\"\n    cluster_password=\"admin\"\n
cluster_check_interval_ms=900\n    cluster_check_status_frequency=1000\n
cluster_mysql_query_rules_save_to_disk=false\n   
cluster_mysql_servers_save_to_disk=false\n    cluster_mysql_users_save_to_disk=true\n   
cluster_proxysql_servers_save_to_disk=false\n   
cluster_mysql_query_rules_diffs_before_sync=2\n   
cluster_mysql_servers_diffs_before_sync=3\n   
cluster_mysql_users_diffs_before_sync=4\n    
cluster_proxysql_servers_diffs_before_sync=30\n\n}\n\nmysql_variables = \n{\n    
monitor_username=\"monitoruser\"\n    monitor_connect_timeout=1000\n    
monitor_password=\"monitorpass\"\n    max_connections=32000\n    
default_charset=utf8\n\n}\n\nproxysql_servers = \n(\n    {\n    hostname=\"192.168.1.10\"\n    
port=6032\n    comment=\"proxy01\"\n    },\n    {\n    hostname=\"192.168.1.11\"\n    
port=6032\n    comment=\"proxy02\"\n    },\n\n)\n\nmysql_servers = \n(\n    { 
address=\"192.168.1.12\" , port=3306, hostgroup=100, max_connections=32000 },\n    { 
address=\"192.168.1.13\" , port=3306, hostgroup=200, max_connections=32000 },\n    { 
address=\"192.168.1.14\" , port=3306, hostgroup=300, max_connections=32000 
},\n\n)\n\nmysql_galera_hostgroups =\n(\n    {\n        writer_hostgroup=100\n        
reader_hostgroup=200\n        backup_writer_hostgroup=300\n        
offline_hostgroup=400\n        writer_is_also_reader=0\n        
max_transactions_behind=1000\n        active=1\n    }\n)\n\nmysql_query_rules =\n(\n    {\n        
rule_id=1\n        active=0\n        match_pattern=\"^SELECT\"\n        
destination_hostgroup=100\n        apply=0\n    },\n    {\n        rule_id=2\n        active=1\n        
match_pattern=\"^UPDATE\"\n        destination_hostgroup=200\n        apply=1\n    },\n    {\n        
rule_id=3\n        active=1\n        match_pattern=\"^FOR .*\"\n        
destination_hostgroup=300\n        apply=1\n    },\n    {\n        rule_id=4\n        active=1\n        
match_pattern=\"SHOW\"\n        destination_hostgroup=400\n        apply=0\n    },\n    {\n        
rule_id=5\n        active=0\n        match_pattern=\"PTFS\"\n        
destination_hostgroup=200\n        apply=1\n    },\n\n)",
```

The task is a typical call to `file`:

```json
{
    "dest": "/etc/proxysql.cnf",
    "follow": false,
    "group": "proxysql",
    "mode": "0600",
    "module_args": {
        "dest": "/etc/proxysql.cnf",
        "follow": false,
        "group": "proxysql",
        "mode": "0640",
        "owner": "proxysql",
        "src": "/home/userx/.ansible/tmp/ansible-local-753936wt440r4/tmpuavewc3y/proxysql.cnf.j2"
    },
    "owner": "root",
    "src": "/home/userx/.ansible/tmp/ansible-local-753936wt440r4/tmpuavewc3y/proxysql.cnf.j2"
}
```

For the code part, what do you think would be the best way to go about it? Maybe we can catch the unintended behaviour by searching for escape characters in the output.

dmsimard: I realize now that the support for rendering diffs in the UI only landed in 1.5.5 which actually hasn't been released yet (see [changelog](https://github.com/ansible-community/ara/releases/tag/1.5.5rc1))

Could you test with the latest release candidate (1.5.5rc1, via ``pip install --pre ara[server]``) and let me know how the format looks to you ? I... forgot to actually release it.

Here's an example of file module diff that I found on the demo which seems to render well: https://demo.recordsansible.org/results/436039.html

dmsimard: Heads up if you are upgrading from 1.4.3 there's two SQL migrations (see upgrade notes in https://ara.readthedocs.io/en/latest/changelog-release-notes.html)

Thulium-Drake: @dmsimard I ran the update to my ARA install to 1.5.5.0 (I also made another PR to the collection, as the PyPi task I wrote never updated the installation ;-) ) and it works like a charm! The diffs are actually readable now!

Thanks :sunglasses:

aorfanos: I can confirm it works for me too using `ara 1.5.5.0rc2.dev6` :) See you in another PR @dmsimard 💯

PR #210, opened 2021-03-01
zuul: set project merge mode to squash-merge

dmsimard:
So Zuul doesn't end up merging every commit from a PR individually but
rather squash them.


dmsimard:
recheck

Issue #220, opened 2021-03-13
API: Add CLI args into playbook parent references


dmsimard:
## What is the idea ?

CLI arguments are available on ``/api/v1/playbooks`` as well as ``/api/v1/playbooks/<id>``, for example:
```json
       {
            "id": 2937,
            "items": {
                "plays": 1,
                "tasks": 26,
                "results": 26,
                "hosts": 1,
                "files": 3,
                "records": 0
            },
            "arguments": {    # <--- this
                "version": null,
                "verbosity": 0,
                "ask_pass": false,
                "private_key_file": null,
```

 but they aren't when viewing a resource such as a host, for example ``/api/v1/hosts/<id>``:

```json
{
    "id": 5,
    "playbook": {
        "id": 5,
        "items": {
            "plays": 1,
            "tasks": 261,
            "results": 214,
            "hosts": 1,
            "files": 34,
            "records": 0
        },
        "labels": [],
        "started": "2020-01-31T15:46:10.232000Z",
        "ended": "2020-01-31T15:48:29.812000Z",
        "duration": "00:02:19.580315",
        "name": null,
        "ansible_version": "2.8.8",
        "status": "completed",
        "path": "/home/zuul/src/opendev.org/openstack/openstack-ansible/tests/bootstrap-aio.yml",
        "controller": "localhost"
        # <--- missing "arguments" key
    },
    "facts": {
```

It isn't particularly heavy and it is rather fixed width so we should include it for consistency and so we can display playbook context (such as a playbook card/summary at the top of a page).

I think we just need to remove the exclude here and it should be enough: https://github.com/ansible-community/ara/blob/06851d8438ad43af7c1cceaecf5c407392966ee6/ara/api/serializers.py#L87-L92

PR #221, opened 2021-03-14
WIP: UI: Transition to bootstrap CSS and misc improvements

dmsimard:
I have no problems with patternfly (and patternfly-react) in ara-web but in my experience, it has been complicated to do raw HTML with patternfly and it appears better suited for javascript interfaces.

I've initially experimented with replacing the CSS with bootstrap while vendoring jquery to see what it would look like and it has been simple and great to work with so I continued the experiment and here we are.

I would had liked to be able to split this work across individual commits and PRs but this will need to land all at once due to the scale of the changes. Edit: the scope has grown a bit and I'll split where it is possible and makes sense.

You're welcome to test these changes locally and provide feedback by checking out the PR branch and running the server as you usually would (i.e, with ``ara-manage runserver``).

In the meantime, I've added a few preview screenshots and GIFs here: https://twitter.com/RecordsAnsible/status/1370210356279267328

Edit: for the time being I've put this up on https://dev.demo.recordsansible.org/, you can compare with https://demo.recordsansible.org/ -- it's the same database underneath.


dmsimard:
TODO: The CLI arguments modal doesn't work when clicked from file, result or host details because they're missing from the API, see: https://github.com/ansible-community/ara/issues/220

dmsimard:
re: mobile view, I've given it a first shot at it via https://github.com/ansible-community/ara/pull/221/commits/cabadb0b7ffa3d1399bec0aea2887998727b94d4

To be clear, I'm not going to spend countless hours optimizing mobile for now but if it can at least be usable without too much effort it's worth looking into. It won't be perfect the first time around.

The implementation in bootstrap and patternfly is different -- it looks different too. I suppose it is a matter of taste and preference -- I think I prefer the bootstrap version ?

patternfly "folds" the table into a mobile-optimized view using data-labels set on cells: https://www.patternfly.org/v4/components/table/design-guidelines/#tables-on-mobile

It looks like this:
Peek 2021-03-14 01-13

while bootstrap leaves the table alone but allows for horizontal scrolling if need be: https://getbootstrap.com/docs/4.5/content/tables/#responsive-tables

Peek 2021-03-14 01-16

hille721:
I tested the new UI today, I love the bootstrap-darkly theme! 

Bust just saw one improvement:  if your are running a playbook on a few hosts (I just run a playbook on over 3000 hosts), then the playbook page is getting quite long (check the scrolling bar on the right):
image

In the current UI, the host section in the playbook page was collapsed when you opened a playbook page with more hosts. But if you expanded it then you had the same problem. In my opinion the best woul be to have a pagination on the host section as you have it already on the task result section.

But the performance increasement is huge from the new UI! With the current one it tooks about 10 seconds to load a playbook page with 3070 hosts (/playbooks/4.html#hosts ), with the new one, it is just about 4 seconds!

dmsimard:
o/ @hille721 thanks for the feedback!

> I love the bootstrap-darkly theme!

Me too ! I think it looks surprisingly good and the green accent goes well with the logo :p
I expect that it may not suit the tastes of everyone so the ability to use the light theme needed to exist, though.

> Bust just saw one improvement: if your are running a playbook on a few hosts (I just run a playbook on over 3000 hosts), then the playbook page is getting quite long (check the scrolling bar on the right)

I'll go on a tangent here and ask about the width.
I've been "benchmarking" or testing this so it renders well in both 1080p and 1440p but I guess wider than that leads to so much whitespace ! What resolution are you running ?

I would like to avoid constraining the width of the pages like GitHub or Twitter to a fraction (say, 800px) so we can make use of that width with all the data that we need to display but we might need to put some max sizes somewhere so it doesn't get out of hand.

> In the current UI, the host section in the playbook page was collapsed when you opened a playbook page with more hosts. But if you expanded it then you had the same problem. In my opinion the best woul be to have a pagination on the host section as you have it already on the task result section.

How it currently works is actually rather naive, if there is a search query the hosts panel will be folded:
https://github.com/ansible-community/ara/blob/a14c808b412df93cf7fe34800603263bf8305a12/ara/ui/templates/playbook.html#L28

I realize now that I didn't do the same thing in this PR (thanks for catching that) but I'd like to find a better way.
Maybe there could be a max height on the cards before making their contents scrollable ?
Otherwise, we know how many hosts there are in the playbook ahead of time (via ``playbook.items.hosts``) so maybe we could fold the card if there are more than X number of hosts ? I'll think about it but I'm open to suggestions :)

I don't have a playbook with 3000 hosts handy but there's this one with 100 hosts where I see an extent of the issue: https://dev.demo.recordsansible.org/playbooks/2895.html

While the results are paginated, the current implementation doesn't allow for the hosts to be paginated easily in addition to the results. It's definitely something we should do in the future but out of scope for now.

> But the performance increasement is huge from the new UI! With the current one it tooks about 10 seconds to load a playbook page with 3070 hosts (/playbooks/4.html#hosts ), with the new one, it is just about 4 seconds!

That's surprising to me because I had no performance expectations or considerations when I started working on this.
Could you share what your browser inspector is saying is taking so long ?

Comparing https://demo.recordsansible.org to https://dev.demo.recordsansible.org I don't see much differences:

Old:
Screenshot from 2021-03-15 17-03-12

New:
Screenshot from 2021-03-15 17-03-35

hille721:
> I'll go on a tangent here and ask about the width.
I've been "benchmarking" or testing this so it renders well in both 1080p and 1440p but I guess wider than that leads to so much whitespace ! What resolution are you running ?

sorry. my screenshot was with a browser zoom level of about 5% percent, that's probably why there is so much whitespace. I just wanted to highlight, that you have to scroll a lot if you have more than ~ 20 hosts and in my case I had a lot of more. But your example with the 100 hosts is also quite well to show that. 

> I realize now that I didn't do the same thing in this PR (thanks for catching that) but I'd like to find a better way.
Maybe there could be a max height on the cards before making their contents scrollable ?
Otherwise, we know how many hosts there are in the playbook ahead of time (via playbook.items.hosts) so maybe we could fold the card if there are more than X number of hosts ? I'll think about it but I'm open to suggestions :)

 l like the idea with the max hosts. Just thought about it, I would say everything more than 10 hosts is too much. because if you do not see directly (or after a bit scrolling) that there is another section on the bottom you will not scroll there. 
 One solution might be also if the task result section would be on the top, and the three columns "hosts, files, records" on the bottom.
 For a lot of hosts a host filter in the host section would be also good. Also the host-id filter in the task results is not really helpful because from where you should know the Ara internal host id. A search field of the host name would be better in my opinion. 
 
 > That's surprising to me because I had no performance expectations or considerations when I started working on this.
Could you share what your browser inspector is saying is taking so long ?

will do that tomorrow morning if I am at work again :)

hille721:
here the comparison:

OLD:
image

NEW:
image

It seems that the old playbook index html was larger than the new ones.

dmsimard:
Nice and unexpected, but I'll take it!

From what I can tell, the old css file takes longer to load and we also no longer need to load an ara.css as well as font files.

It accounts for 2MB less in your screenshot. Not bad!

dmsimard:
@hille721 I checked what it looked like just by adding ``max-height: 300px; overflow-y: auto;`` to the hosts card and it's not bad.
I think I could live with that until we implement proper pagination.

On desktop:
Peek 2021-03-16 13-17

and mobile:
Peek 2021-03-16 13-20

dmsimard:
> For a lot of hosts a host filter in the host section would be also good. Also the host-id filter in the task results is not really helpful because from where you should know the Ara internal host id. A search field of the host name would be better in my opinion.

Yeah I realize the host and task IDs are not "human readable", the way they're used right now is if you click on a task name it'll filter for that task id and if you click on a hostname (or one of it's statuses), it'll filter for that host (and perhaps status).

The way it works is basically a mashup of three tables:
- tasks
- hosts
- results

We first get all the results of a playbook, roughly the equivalent of ``/api/v1/results?playbook=<id>``):
https://github.com/ansible-community/ara/blob/a14c808b412df93cf7fe34800603263bf8305a12/ara/ui/views.py#L80-L81

Results are an association of a task and a host and we "backfill" detailed information about tasks and hosts here:
https://github.com/ansible-community/ara/blob/a14c808b412df93cf7fe34800603263bf8305a12/ara/ui/views.py#L89-L93

In other words, the result references (by foreign key) the task id and the host id (not their names!) and I haven't yet found how to make that possible. I haven't given this much thought yet but maybe under the hood we could search for a hostname for that playbook (``/api/v1/hosts?playbook=<id>&name=foo``), get it's ID and then filter with that -- but strictly API speaking, we cannot search a result by hostname right now.

hille721:
> In other words, the result references (by foreign key) the task id and the host id (not their names!) and I haven't yet found how to make that possible.

I had exactly such a problem in a project at work some weeks ago and if I remeber correctly I solved it with a [SlugRelatedField](https://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield) in the serializer. 

Thus it should be simple by changing the  PrimaryKeyRelatedFields to SlugRelatedFields.

dmsimard:
I made the cards scrollable in https://github.com/ansible-community/ara/pull/221/commits/90550b32393d16a758457ef776ac2fa9e09cc6bf

It's an improvement over what we had, thanks for the feedback :D

We had this issue with files as well for larger playbooks, for example before:
Selection_101

and after:
Selection_102

dmsimard:
> I had exactly such a problem in a project at work some weeks ago and if I remeber correctly I solved it with a SlugRelatedField in the serializer.
>
> Thus it should be done by changing the PrimaryKeyRelatedFields to SlugRelatedFields in the List*Serializers.
>
> This would have been also the benefit to be able to easier use the API from external sources, e.g. checking for failed Ansible tasks of a host in a monitoring system would be able with one ara API call.

It's cool to know that it's possible \o/
For info we already have a custom SlugRelatedField here for labels: https://github.com/ansible-community/ara/blob/a14c808b412df93cf7fe34800603263bf8305a12/ara/api/fields.py#L79-L83

dmsimard:
The last big item the to-do list before I think this is ready to land for a first iteration is ensuring that the static generation still works and I'll tackle that next.

The idea is to generate a static copy of the webapp "manually" (via https://github.com/ansible-community/ara/blob/master/ara/ui/management/commands/generate.py) without searching, ordering, filtering and using relative links.

This is why before this rework, URLs that could be "dynamic" used the ``{% url %}`` macro but some used links like ``../playbooks/<id>.html`` like here for example:
https://github.com/ansible-community/ara/blob/a14c808b412df93cf7fe34800603263bf8305a12/ara/ui/templates/base.html#L40

Using ``../`` links allows the generated content to be hosted under any directory (like http://logserver/some/directory/ara-report) while using the url macro expects everything to be under ``/`` which wouldn't work since, for example, the css lives at ``logserver/some/directory/ara-report/static/css`` not at ``logserver/static/css``.

There are some additional ``{% if static_generation %}`` conditions sprinkled around the templates to handle these cases or exclude dynamic content such as searching and sorting/ordering.

I'll try and find out if it's possible to do something cleaner.

ara/ui/static/css/bootstrap-darkly.min.css
```diff
@@ -0,0 +1,12 @@
...
...lor:#fff}.nav-pills .nav-item.open .nav-link,.nav-pills .nav-item.open .nav-link:focus,.nav-pills .nav-item.open .nav-link:hover,.nav-pills .nav-link,.nav-pills .nav-link.active,.nav-pills .nav-link.active:focus,.nav-pills .nav-link.active:hover,.nav-tabs .nav-item.open .nav-link,.nav-tabs .nav-item.open .nav-link:focus,.nav-tabs .nav-item.open .nav-link:hover,.nav-tabs .nav-link,.nav-tabs .nav-link.active,.nav-tabs .nav-link.active:focus,.nav-tabs .nav-link.active:hover{color:#fff}.breadcrumb a{color:#fff}.pagination a:hover{text-decoration:none}.close{opacity:.4}.close:focus,.close:hover{opacity:1}.alert{border:none;color:#fff}.alert .alert-link,.alert a{color:#fff;text-decoration:underline}.alert-primary{background-color:#375a7f}.alert-secondary{background-color:#444}.alert-success{background-color:#00bc8c}.alert-info{background-color:#3498db}.alert-warning{background-color:#f39c12}.alert-danger{background-color:#e74c3c}.alert-light{background-color:#adb5bd}.alert-dark{background-color:#303030}.list-group-item-action{color:#fff}.list-group-item-action:focus,.list-group-item-action:hover{background-color:#444;color:#fff}.list-group-item-action .list-group-item-heading{color:#fff}
```
dmsimard, line 12: TODO: just noticed this: ``/@import url(https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,400;0,700;1,400&display=swap);``. I'd much rather have the font self hosted and it doesn't seem too complicated according to this issue: https://github.com/thomaspark/bootswatch/issues/573

dmsimard: Just removing the import seems to be sufficient without any discernable impact that I can see -- the font appears to be the same and we're not calling out to google anymore.

Before:
Screenshot from 2021-03-17 14-38-00

After:
Screenshot from 2021-03-17 14-37-14

dmsimard:
Ok I think I got the static generation working properly and I removed the font import from the darkly theme that I spotted.

I'll clean this up and will open a new PR before closing this one.

Thanks for your feedback while I was working on this @hille721 :+1:

dmsimard:
Cleaned this up and opened #225. Closing this one.

PR #222, opened 2021-03-15
api: don't exclude arguments from playbook parent references

dmsimard:
hosts, plays, tasks, results, files and records have references to
their parent playbook in their detailed view (i.e, ``/api/v1/hosts/<id>``)
but until now it didn't include arguments.

It turns out it would be convenient to include the arguments in this
reference so we can have it readily available for providing context.

The arguments are relatively small so it should be fine to make them
available without incurring significant overhead.

Fixes: https://github.com/ansible-community/ara/issues/220


dmsimard:
recheck

PR #231, opened 2021-03-24
zuul: remove squash-merge mode

dmsimard:
It doesn't work the way I thought it did. It discards all commits and
their messages, leaving only the PR subject and PR number.


dmsimard:
recheck

PR #232, opened 2021-03-26
Move container-fluid class one level higher

guillaumevincent:
There are block alignment errors that distract users.
By putting container-fluid class one level higher, we avoid oversights in the child blocks.
And we improve the alignment of the blocks.


dmsimard:
Thanks !

PR #233, opened 2021-03-26
Fix pagination margin

guillaumevincent:
Visually to keep the rhythm in the page, the spaces must be consistent.
This PR is there to put the same space above and below the pagination.


dmsimard:
recheck

dmsimard:
This is indeed better, will merge after CI is green. Thanks !

PR #234, opened 2021-03-26
Fix vertical alignment of the search button

guillaumevincent:
You use a h5 inside a button and this is probably not good in term of accessibility.
An other problem is that h5 introduce a little bottom margin, so the text is not align vertically.
If you want to make the text bigger in a button, you can use the btn-lg class modifier.
This modifier fix the problem of the bottom margin also.


dmsimard:
Much better :)

PR #235, opened 2021-03-26
Make playbook status icon clickable

guillaumevincent:
Like tags we can't filter playbooks by clicking on the status icon directly.
This patch move status icon in his own file

Status icon in filters are not badges anymore. Badge looks like button and user will be confuse to see a button next a checkbox. This is a visual preference to avoid surcharging the interface:

Before 
image
After
image

Tell me if you prefer to keep the badge.

This patch also fix the size of the control on a playbook line to make it consistent:

Before
image
After
image


dmsimard:
I like it, good idea on making the status clickable. Thanks !

PR #238, opened 2021-03-26
ui: try a different color scheme for labels

dmsimard:
Before:
112403654-3173bc80-8ce5-11eb-916c-d7d1ce1f06ef

After:
Selection_130


dmsimard:
How about green instead ?

Selection_133

Selection_134

Also, maybe it's a bit optimistic to display a lot of labels in a narrow column, maybe we should display labels another way ?

In the playbook card we have them on another line which can make use of the entire width:
Selection_136

Very open to suggestions :)

dmsimard:
recheck

PR #246, opened 2021-04-02
ui: Fix header for CLI args in playbook card

dmsimard:
0c7d8424187dfa9173f5a50c95ab930f2fcb0704 fixed the header when viewing
CLI args from the playbook index but the one from the playbook card was
forgotten.


dmsimard:
recheck

Issue #218, opened 2021-03-12
adding a possibilty to search for hosts on the starting page


hille721:
## What is the idea ?
It would be greate to have a possibilty to search for a host in the web ui of the api server:

![image](https://user-images.githubusercontent.com/38468371/110927585-1e94db80-8326-11eb-9fe1-564b1a189ea9.png)

The result shoud be then all playbook runs with this host.

dmsimard:
Hi @hille721 and thanks for the issue.

I have a WIP PR for this though I do not have an ETA yet: https://github.com/ansible-community/ara/pull/206

In the meantime, this is already implemented in the CLI with the ``ara host list`` command: https://ara.readthedocs.io/en/latest/cli.html#ara-host-list

There is a ``name`` argument which lets you search for specific hosts.

hille721:
Hi @dmsimard,

THAT IS AWESOME! :+1: 

just installed your PR branch and tested it, definitive more as I expected, but I love it! 
Is there anything where I can help you to finish it? Testing, checking your implementation, implement stuff, writinge documentation, ...?
Just let me know :-)

hille721:
Hi @dmsimard,

THAT IS AWESOME! :+1: 

just installed your PR branch and tested it, definitive more as I expected, but I love it! 
Is there anything where I can help you to finish it? Testing, checking your implementation, implement stuff, writinge documentation, ...?
Just let me know :-)

dmsimard:
Thanks for checking it out and appreciate the offer for help !

This is how I see things eventually playing out:
- The index of the built-in UI is the list of playbooks, analogous to ``ara playbook list`` on the CLI or [/api/v1/playbooks](https://demo.recordsansible.org/api/v1/playbooks) on the API
- The hosts page (i.e, the PR) would be analogous to ``ara host list`` and [/api/v1/hosts](https://demo.recordsansible.org/api/v1/hosts) for the API

Then, following the same pattern, we could have similar pages for plays, tasks and results.

For hosts specifically, I'm not sure I want to go there for a first implementation but I think there are opportunities for doing things like [ansible-cmdb](https://github.com/fboender/ansible-cmdb) since we also have host facts when they are gathered:
![screenshot-overview](https://user-images.githubusercontent.com/1291204/110987737-8acf0980-833d-11eb-843e-dddcb43fbae3.png)

For now the PR is almost ready to land and I can probably take care of it soon.
[...]

Issue #153, opened 2020-08-04
database: consider jsonfield with django >= 3.1


dmsimard:
## What component is this about ?

database model, potentially requiring API changes

## What is the idea ?

We are not ready to upgrade Django yet (waiting for the next LTS) but django 3.1 was just released with jsonfield support for all database backends whereas it was only compatible with postgresql before: https://docs.djangoproject.com/en/3.1/releases/3.1/#jsonfield-for-all-supported-database-backends

Since a lot of things from Ansible is JSON, this could help us simplify the database schema and also make the fields easily searchable -- I'm thinking like searching through results, host facts or CLI arguments, for example.

dmsimard:
3.2 LTS is out: https://www.djangoproject.com/weblog/2021/apr/06/django-32-released/

dmsimard:
WIP upgrade to django 3.2: https://github.com/ansible-community/ara/pull/360

dmsimard:
By the way, documentation from django and django-rest-framework on jsonfield:
- https://docs.djangoproject.com/en/3.2/ref/models/fields/#jsonfield
- https://www.django-rest-framework.org/api-guide/fields/#jsonfield
- https://www.django-rest-framework.org/api-guide/filtering/#searchfilter

PR #247, opened 2021-04-04
Distinct hosts

hille721:
Hi @dmsimard, 

I had another idea how to solve the "distinct hosts problem" (compare discussion here #218), with this WIP PR I want to share my idea and open the discussion:

The idea is to create a new model `DistinctHost` with basically only two fields:
* `name`: the hostname, which will be the primary key of the database table
* `latest_host`: which is a `ForeignKey` to the `Host` model

Thus in this model we have alway a list with all distinct hosts and directly the relation to the latest `Host` object. 

The trick now is to update the `DistinctHost` model if a new `Host` model entry will be created. We could do that already by overwriting the `save` method of the `Host` model, but I think that's not a best practice. And because in our case we only create `Host` entries via the API, I implemented a `perform_create` method on the `HostViewSet`, which updates (`update_or_create`) the `DistinctHost` model via an `DistinctHostSerializer`.

For development and testing purposes I've also already added a API endpoint `/api/v1/distincthosts` with a ListSerializer and the enpoints `/api/v1/distincthosts/HOSTNAME` with a DetailedSerializer. 
Surely there could be still some improvements, but for first tests it should be enough.

I haven't yet tested the performance in scale, but I do not see a reason why it should be slow. 

The advantage of my approach here, is that we do not touch existing model, but only introduce a new one. But with that we could now implement all the things we thought about in #218. Furthermore I see also an advantage in having an API endpoint with which you can directly check the last result for a host (`/api/v1/distincthosts/HOSTNAME`).


dmsimard:
@hille721 thanks for working on this! I don't have an opinion yet but will try to reach out to people that might. I'd like to focus getting 1.5.6 out the door first and will circle back to this afterwards.

dmsimard:
o/ apologies for the momentary hiatus, I am still very much interested in the feature and will review.

ara/api/views.py
```diff
@@ -104,6 +104,29 @@ def get_serializer_class(self):
...
             return serializers.HostSerializer
 
+    def perform_create(self, serializer):
+        """ create or update DistinctHost object """
+
+        instance = serializer.save()
+        distinct_host = serializers.DistinctHostSerializer(data={"name": instance.name, "latest_host": instance.id})
+        if distinct_host.is_valid():
+            distinct_host.save()
+        # TODO: exception handling if not valid
```
dmsimard, line 114: This seems to do something similar to what we are doing with the serializer:
https://github.com/ansible-community/ara/blob/31f090e4119d1ac55a9807b9e605082cdd3db1b9/ara/api/serializers.py#L311-L331

This makes me wonder (not a django expert): could we update the latest host via the serializer instead ? Or in other words, should we move the part from the serializer to the view instead ?

dmsimard:
Looks good at first glance, I'll test it a bit and see what https://github.com/ansible-community/ara/pull/206 might look like rebased on top of this.

ara/api/migrations/0009_distincthost.py
```diff
@@ -0,0 +1,26 @@
...
+                ('name', models.CharField(max_length=255, primary_key=True, serialize=False)),
+                ('created', models.DateTimeField(auto_now_add=True)),
+                ('updated', models.DateTimeField(auto_now=True)),
+                ('latest_host', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='api.Host')),
+            ],
+            options={
+                'db_table': 'distinct_hosts',
+            },
+        ),
+    ]
```
dmsimard, line 26: This create the DistinctHost model -- maybe we could also add a migration step that "computes" the current distinct hosts and inserts the rows ? That way, after the migration is done, the list is already populated and ready to be used.

We have some python code running in this migration as an example: https://github.com/ansible-community/ara/blob/master/ara/api/migrations/0004_duration_in_database.py

ara/api/models.py
```diff
@@ -240,6 +240,27 @@ def __str__(self):
         return "<Host %s:%s>" % (self.id, self.name)
 
 
+class DistinctHost(models.Model):
+    """
+    Distinct list of hosts with latest `Host` object id related.
+
+    We can not inherit from Base because we won't to use `name` as primary key.
```
dmsimard, line 247: :+1: 
```suggestion
    We can not inherit from Base because we want to use `name` as primary key.
```

ara/api/models.py
```diff
@@ -240,6 +240,27 @@ def __str__(self):
...
+    """
+
+    class Meta:
+        db_table = "distinct_hosts"
+
+    name = models.CharField(max_length=255, primary_key=True)
+    latest_host = models.ForeignKey(Host, on_delete=models.CASCADE)
+
+    # because we can't inherit from Base (see above) we have to define this here additionally)
```
dmsimard, line 256:
```suggestion
    # because we can't inherit from Base (see above) we have to define this here additionally
```

dmsimard, line 256:
```suggestion
    # because we can't inherit from Base (see above) we have to define this here additionally
```

ara/api/migrations/0009_distincthost.py
dmsimard: I was curious and tried something that seems to work: https://gist.github.com/dmsimard/d4fea5ce3aff720f706270d37b9632d4

dmsimard:
> Looks good at first glance, I'll test it a bit and see what #206 might look like rebased on top of this.

I've got a very rough WIP rebase of the host index patch I had (had to carry it through the UI refactor, it was using the old HTML/CSS, whew) but it mostly "just worked" on top of your latest_host implementation with the sql migration I provided in this [gist](https://gist.github.com/dmsimard/d4fea5ce3aff720f706270d37b9632d4):
Screenshot from 2021-06-18 00-04-41

If you're curious, I've put up a testing branch [here](https://github.com/dmsimard/ara/tree/wip_host_index) and I've deployed it to https://dev.demo.recordsansible.org/hosts -- it uses a copy of the MySQL database from demo.recordsansible.org with the migration run on it.

hille721:
Hi @dmsimard,
saw your comments shortly before I went to bed. And wow, what have you already achieved during I slept 👍 

I had also no time the last weeks to finaly also do a performance test with thousands of hosts, but now I'm motivated again.

ara/api/views.py
hille721: yes you are right, we could do it in the `create` method of the `serializers.py` as well. Honestly I don't know anymore why I did it in the  `perform_create` method of the `views.py`.
I'm also not a "Django Expert" and thus also not sure what's the best practice here, but I will try to find out.

But it definitely make sense to make it consistent in this project.

ara/api/migrations/0009_distincthost.py
hille721: This is an awesome approach, I haven't thought about that, but really like it. 

Regarding your question:
> TODO: Should we sort by another column ?

I would say the updated column is fine because this will be also the criteria for future data (as it is implemented right now).

hille721: I've added you proposal as well to this PR branch

dmsimard:
> I had also no time the last weeks to finaly also do a performance test with thousands of hosts, but now I'm motivated again.

I didn't time it but we could benchmark before release. The migration completed near instantly for the demo database which has ~8000 hosts.

I was pleasantly surprised because I expected it to take longer than that. The migration is written to only iterate over the hosts once and quickly skips host names that have been processed so I guess it pays off.

hille721:
I've testet it now with 10,000 hosts (~3050 distinct hosts). 
It's too fast to benchmarkt it :) 

Migration was fast and also the `/api/v1/distincthosts` has now the same reaction time as `/api/v1/hosts`.

ara/api/models.py
```diff
@@ -240,6 +240,27 @@ def __str__(self):
...
+    """
+    Distinct list of hosts with latest `Host` object id related.
+
+    We can not inherit from Base because we want to use `name` as primary key.
+    """
+
+    class Meta:
+        db_table = "distinct_hosts"
+
+    name = models.CharField(max_length=255, primary_key=True)
```
dmsimard, line 253: From an implementation standpoint, I was wondering whether we should use the name as the primary key or keep the usual integer and have a name field (with UNIQUE) like we do for labels for example: https://github.com/ansible-community/ara/blob/31f090e4119d1ac55a9807b9e605082cdd3db1b9/ara/api/models.py#L69

Having the name as a primary key provides the ability to do something like: ``/api/v1/distincthosts/<name>`` (rather than ``<id>``) which is nifty. This avoids needing to search for it like ``/api/v1/distincthosts?name=<name>`` and then recover the ID from the list, though the name search is partial so it could potentially return more than one result if the name search is a subset of another host name.

More importantly, it doesn't prevent us from searching by name anyway -- i.e, https://dev.demo.recordsansible.org/hosts?name=localhost returns localhost in addition to localhost1, localhost2, etc.

With that in mind, it's probably a good idea to keep name as the primary key.

ara/api/views.py
hille721: With https://github.com/ansible-community/ara/pull/247/commits/cc564107a17dd659714f95f11b9d7fc3ed1fa0df I moved now the creation/update of the DistinctHost entry to the `create` method of the `HostSerializer` in `serializers.py`.

This is basicly exactly how it is descrived in the official DRF guide:
https://www.django-rest-framework.org/api-guide/relations/#writable-nested-serializers

dmsimard:
Some braindump high level considerations to help think ahead and test the implementation:
- UI: I think it makes sense to expose the new WIP host index at ``/hosts`` because host details end up at ``/hosts/<id>.html``.
- UI: We should probably have a form checkbox (default to latest hosts?) that allows to switch from the full list of hosts (i.e, ``/api/v1/hosts``) to only the latest ones (``/api/v1/distincthosts``). I can see a use case for showing all the hosts, especially for finding all the playbook runs for a particular host instead of only the latest one.
- UI: I believe we'll eventually want to move what we currently call the index (list of playbook reports) to ``/playbooks``, if not for consistency with ``/api/v1/playbooks``, it would be for consistency with ``/hosts``, assuming that's where we end up putting the new host index.

dmsimard:
@hille721 btw, it was wise not to write tests to begin with as to avoid the overhead of changing them if we decide to change the implementation but I think we're in the right direction and we should add a few tests. We can take inspiration from the current host tests which are composed of two main parts:
- https://github.com/ansible-community/ara/blob/master/ara/api/tests/factories.py#L104-L115
- https://github.com/ansible-community/ara/blob/master/ara/api/tests/tests_host.py

We don't need to test /everything/ but a minimum amount of coverage that tells us that things aren't horribly broken would be nice :)

dmsimard:
Ignore the failing container image CI job for now, it's a bit neglected and needs love. I'll take care of it.

ara/api/views.py
dmsimard: Nice! I must have followed that documentation when writing the initial implementation ¯\\_ (ツ)_/¯

hille721:
> * UI: I think it makes sense to expose the new WIP host index at `/hosts` because host details end up at `/hosts/<id>.html`.
> * UI: We should probably have a form checkbox (default to latest hosts?) that allows to switch from the full list of hosts (i.e, `/api/v1/hosts`) to only the latest ones (`/api/v1/distincthosts`). I can see a use case for showing all the hosts, especially for finding all the playbook runs for a particular host instead of only the latest one. **Edit**: We can also implement this in the CLI with a `--latest` flag or something like that.

Yes, we need definitly both, a view with just the latest/ distinct hosts and a view with all runs of a host. 

Idea: How about to have a host view with all distinct hosts, similar to your screenshot above, but maybe with some facts already in the table (OS, ...) and let's call the date column "last report" and the Task results column "Last Result".
And istead of only a link to the Last Report, we put in a link to a view with all results of that host (thus API speaking `/api/v1/hosts?name=HOSTNAME`)

To make us a bite more flexible in the UI, I've now also added a new field `hosts` to the `DistinctHost` model where we have diretly all related `Host` model entries for one `DistinctHost` and not only the latest one:
image

I'm not sure if this is really necessary, because we woul also get all related `Host` entries via `/api/v1/hosts?name=HOSTNAME`, but maybe it is in some situations easier to directly having this relation and also this could maybe save some calculation effort.

dmsimard:
> > * UI: I think it makes sense to expose the new WIP host index at `/hosts` because host details end up at `/hosts/<id>.html`.
> > * UI: We should probably have a form checkbox (default to latest hosts?) that allows to switch from the full list of hosts (i.e, `/api/v1/hosts`) to only the latest ones (`/api/v1/distincthosts`). I can see a use case for showing all the hosts, especially for finding all the playbook runs for a particular host instead of only the latest one. **Edit**: We can also implement this in the CLI with a `--latest` flag or something like that.
> 
> Yes, we need definitly both, a view with just the latest/ distinct hosts and a view with all runs of a host.

To be determined but I feel like we can achieve this with a single view in the UI and just change the queryset based on whether the user wants to see only the latest hosts or not (with i.e, a checkbox).

> Idea: How about to have a host view with all distinct hosts, similar to your screenshot above, but maybe with some facts already in the table (OS, ...)

:+1: I think there is an opportunity to do something similar to what [ansible-cmdb looks like](https://github.com/fboender/ansible-cmdb#about), though I don't know what it would look like yet. I'd like to land a first "ordinary" iteration that doesn't include host facts and we can work on adding them later.

> and let's call the date column "last report" and the Task results column "Last Result".

:+1:, though if we re-use the same view we might want to have conditional column names -- last report/result is appropriate when viewing the latest hosts but not so much when viewing all hosts.

> And istead of only a link to the Last Report, we put in a link to a view with all results of that host (thus API speaking `/api/v1/hosts?name=HOSTNAME`)

Yeah, maybe we could link to the same view but clear/uncheck the "latest" checkbox and filter by name so we see all reports for that name.

> To make us a bite more flexible in the UI, I've now also added a new field `hosts` to the `DistinctHost` model where we have diretly all related `Host` model entries for one `DistinctHost` and not only the latest one:
> image
> 

Note: you blanked the first name but not the second one

> I'm not sure if this is really necessary, because we woul also get all related `Host` entries via `/api/v1/hosts?name=HOSTNAME`, but maybe it is in some situations easier to directly having this relation and also this could maybe save some calculation effort.

I'm not sure about this one either, like you said it feels a bit redundant and could even get a bit unwieldy if there are a lot of reports for a given hostname.

However your comment combined with the idea about displaying some host facts, did make me think about something.
We don't have the host facts in the distincthosts view:
Screenshot from 2021-06-18 11-09-40

Host facts are currently only available in the detailed host view (``/api/v1/hosts/<id>``) and aren't included in the list view (similar task results content, file content, etc.), mostly out of a performance concern (compression/decompression, unnecessary serialization, etc).

There is less of a performance concern for ``/api/v1/distincthosts`` since by design, there are less hosts to iterate over and I would be comfortable if we included the host facts so that we can leverage them in the UI.

hille721:
@dmsimard, completly aggree with everything above.

> I think there is an opportunity to do something similar to what ansible-cmdb looks like, though I don't know what it would look like yet. I'd like to land a first "ordinary" iteration that doesn't include host facts and we can work on adding them later.

something as ansible-cmd is also my dream. But yes, let's go with an ordinary iteration first.

> I'm not sure about this one either, like you said it feels a bit redundant and could even get a bit unwieldy if there are a lot of reports for a given hostname.

Yes, I think that's nonsense because redundant. Have deleted it again.

> However your comment combined with the idea about displaying some host facts, did make me think about something.
We don't have the host facts in the distincthosts view

Good catch, it is in now with https://github.com/ansible-community/ara/pull/247/commits/fbc73b0c570aa19133ef23822ce337703d80f859.

I will try to write some test next weeks and then the Backend/API part should be already done.

ara/api/models.py
```diff
@@ -240,6 +240,27 @@ def __str__(self):
...
+    latest_host = models.ForeignKey(Host, on_delete=models.CASCADE)
```
dmsimard, line 254: Just a thought: maybe we should name this ``latest`` instead of ``latest_host`` ?

An example usage in a template:
Screenshot from 2021-06-18 20-09-07

So, for example, ``host.latest_host.id`` would become ``host.latest.id`` and so on.

dmsimard:
Showerthought: maybe we could adopt a similar approach with a ``DistinctPlaybook`` table based on the playbook path instead of the host name.

dmsimard:
> > > * UI: I think it makes sense to expose the new WIP host index at `/hosts` because host details end up at `/hosts/<id>.html`.
> > > * UI: We should probably have a form checkbox (default to latest hosts?) that allows to switch from the full list of hosts (i.e, `/api/v1/hosts`) to only the latest ones (`/api/v1/distincthosts`). I can see a use case for showing all the hosts, especially for finding all the playbook runs for a particular host instead of only the latest one. **Edit**: We can also implement this in the CLI with a `--latest` flag or something like that.
> > 
> > 
> > Yes, we need definitly both, a view with just the latest/ distinct hosts and a view with all runs of a host.
> 
> To be determined but I feel like we can achieve this with a single view in the UI and just change the queryset based on whether the user wants to see only the latest hosts or not (with i.e, a checkbox).

I experimented with a single view for both all hosts and only the latest with a checkbox:
Screenshot from 2021-06-19 15-15-22

I don't like it because it turns out the template and the view needs a bunch of additional logic and conditionals. It's not very pretty, gets complicated and I prefer simple things :p

I will leave out the ``/api/v1/hosts`` version alone for now and will work with ``/api/v1/distincthosts`` instead.

dmsimard:
I've updated https://dev.demo.recordsansible.org/hosts with your changes and experimented with a basic implementation for some host facts:
Screenshot from 2021-06-19 16-25-03

ara/api/models.py
hille721: sure, we can just rename it. Makes definitely sense.

hille721:
> Showerthought: maybe we could adopt a similar approach with a DistinctPlaybook table based on the playbook path instead of the host name.

Honestly I don't see a use case for this. In fact I would have some concerns: The path of a playbook is definitely not a unique descriptor of a playbook. E.g. let's assume a team with a playbook `playbook.yml` in a GIT repository. And everyone in the team is cloning the repo in her/his home and running Ansible from there.Then you will get a DistintPlaybook for every team member.
 Also I see problems if you are running Ansible not directly and e.g. in Rundeck. Lets assume a Rundeck job with an inline Ansible Plabook, you will get an arbitrary temp file as path (something as `/tmp/XXXX`).

hille721:
> I don't like it because it turns out the template and the view needs a bunch of additional logic and conditionals.

where exactly you need those logic and conditionals. Maybe we could make the `DistinctHostSerializer` looking as same as the `HostSerializer`? Because then no conditionals in the view should be necessary?

hille721:
> I've updated https://dev.demo.recordsansible.org/hosts with your changes and experimented with a basic implementation for some host facts

Thus you would introduce facts as kind of labels? For a first iteration fine, but in the end I would love to see some important facts directly as columns, e.g. OS

hille721:
> I've updated https://dev.demo.recordsansible.org/hosts with your changes and experimented with a basic implementation for some host facts

Two more things: If I'm click on the hostname, I would like to see all playbook reports with this host (`/api/v1/hosts?name=HOSTNAME`).
If I click on a fact, I want to filter on that fact. I guess for that we have to introduce a filter for `/api/v1/distincthosts`?

dmsimard:
> > I don't like it because it turns out the template and the view needs a bunch of additional logic and conditionals.
> 
> where exactly you need those logic and conditionals. Maybe we could make the `DistinctHostSerializer` looking as same as the `HostSerializer`? Because then no conditionals in the view should be necessary?

I have thrown away the code by now but the queryset in the view ends up different based on whether or not only the latest playbooks should be displayed (i.e, ``/api/v1/hosts`` -> ``models.Host.objects.all()`` vs ``/api/v1/distincthosts`` -> ``models.DistinctHost.objects.all()``), the serializer is different and then based on whether it's one or the other, the data is located under different attributes.

I don't think we could use the same serializer due to the decision to include facts unless we want to always return facts and even then, the weight of the response at https://dev.demo.recordsansible.org/api/v1/distincthosts makes my browser lag a bit.

dmsimard:
> > I've updated https://dev.demo.recordsansible.org/hosts with your changes and experimented with a basic implementation for some host facts
> 
> Thus you would introduce facts as kind of labels?

They are displayed similarly as labels for lack of better PoC but they're just buttons/links to the fact itself with no search capabilities for now. Last time I've looked at enabling search, it turned out to be pretty difficult due to it being a nested json. Django 3.1 enables [native json fields](https://github.com/ansible-community/ara/issues/153) and I think we have a shot at implementing this once we use them, though it may be "2.0" territory.

> For a first iteration fine, but in the end I would love to see some important facts directly as columns, e.g. OS

The challenge with displaying facts as columns is that you can only have so many until the width becomes an issue, not all hosts have gathered facts and then you need to choose which facts to display. The way it's implemented in javascript for cmdb provides the ability to customize which columns are shown: https://rawgit.com/fboender/ansible-cmdb/master/example/html_fancy.html

I'm not a javascript person but it's something we can consider, while the intent is not necessarily to copy what cmdb does or how it does it :)

dmsimard:
> Two more things: If I'm click on the hostname, I would like to see all playbook reports with this host (`/api/v1/hosts?name=HOSTNAME`).

Yep, this view does not exist yet. Only the distincthost one. It can exist eventually :)

> If I click on a fact, I want to filter on that fact. I guess for that we have to introduce a filter for `/api/v1/distincthosts`?

I don't think this is possible yet (see my response about django 3.1 in https://github.com/ansible-community/ara/pull/247#issuecomment-864481001)

hille721:
> > For a first iteration fine, but in the end I would love to see some important facts directly as columns, e.g. OS
> 
> The challenge with displaying facts as columns is that you can only have so many until the width becomes an issue, not all hosts have gathered facts and then you need to choose which facts to display. The way it's implemented in javascript for cmdb provides the ability to customize which columns are shown: https://rawgit.com/fboender/ansible-cmdb/master/example/html_fancy.html
> 
> I'm not a javascript person but it's something we can consider, while the intent is not necessarily to copy what cmdb does or how it does it :)

Would be definitely a killer feature in the future!

hille721:
> > > I've updated https://dev.demo.recordsansible.org/hosts with your changes and experimented with a basic implementation for some host facts
> > 
> > 
> > Thus you would introduce facts as kind of labels?
> 
> They are displayed similarly as labels for lack of better PoC but they're just buttons/links to the fact itself with no search capabilities for now. Last time I've looked at enabling search, it turned out to be pretty difficult due to it being a nested json. Django 3.1 enables [native json fields](https://github.com/ansible-community/ara/issues/153) and I think we have a shot at implementing this once we use them, though it may be "2.0" territory.



>> If I click on a fact, I want to filter on that fact. I guess for that we have to introduce a filter for /api/v1/distincthosts?
>
> I don't think this is possible yet (see my response about django 3.1 in #247 (comment))

Yes, you can't directly filter in  the queryset, but you could filter afterwards on `serializers.data`:
``` python
response = list(serializer.data)

for host in serializer.data:
    if not host['latest_host']['facts'].get(factname) == factvalue:
        response.remove(host)
```
API call would be then some thing like this: `/api/v1/distincthosts?factname=ansible_distribution&factvalue=Centos`

dmsimard:
> > > If I click on a fact, I want to filter on that fact. I guess for that we have to introduce a filter for /api/v1/distincthosts?
> > I don't think this is possible yet (see my response about django 3.1 in #247 (comment))
> Yes, you can't directly filter in the queryset, but you could filter afterwards on `serializers.data`:
> 
> ```python
> response = list(serializer.data)
> 
> for host in serializer.data:
>     if not host['latest_host']['facts'].get(factname) == factvalue:
>         response.remove(host)
> ```
> 
> API call would be then some thing like this: `/api/v1/distincthosts?factname=ansible_distribution&factvalue=Centos`

That's an interesting approach and one I hadn't considered since I figured that doing the filtering in-database was the way to go performance-wise. Maybe something to consider.

hille721:
> > > > If I click on a fact, I want to filter on that fact. I guess for that we have to introduce a filter for /api/v1/distincthosts?
> > > I don't think this is possible yet (see my response about django 3.1 in #247 (comment))
> > Yes, you can't directly filter in the queryset, but you could filter afterwards on `serializers.data`:
> > 
> > ```python
> > response = list(serializer.data)
> > 
> > for host in serializer.data:
> >     if not host['latest_host']['facts'].get(factname) == factvalue:
> >         response.remove(host)
> > ```
> > 
> > API call would be then some thing like this: `/api/v1/distincthosts?factname=ansible_distribution&factvalue=Centos`
> 
> That's an interesting approach and one I hadn't considered since I figured that doing the filtering in-database was the way to go performance-wise. Maybe something to consider.

sure, in-database filtering is of course the prefered approach.
But in this case I would assume that the overhead of such a simple Python list iteration is negligible compared to the time of getting all hosts.

dmsimard:
I thought about host fact searching/filtering and while I can see it working for simple strings (i.e, ``ansible_os_family``) but there's a lot of facts which are lists or dicts where I wouldn't expect it to work. I guess we would only support it for strings ?

hille721:
> I thought about host fact searching/filtering and while I can see it working for simple strings (i.e, ``ansible_os_family``) but there's a lot of facts which are lists or dicts where I wouldn't expect it to work. I guess we would only support it for strings ?

true. But I guess it will be generally difficult to search for list or dict facts. I don't think that using a `JsonField` in the database will easily solve this problem.

dmsimard:
I spent some time experimenting with host fact filtering yesterday and at least came up with the django_filters that was missing (you had written a [TODO](https://github.com/ansible-community/ara/pull/247/files#diff-637d3b8d855153c8a4278bf51d5d76a772292039f58756bafaab0bead0cd33a3R109)): https://gist.github.com/dmsimard/0da4b011fa84addbfbdc9f4e87e3e387

It works for those fields but I haven't been able to figure out the right implementation for the host facts. I wasn't able to find the right way to plumb your suggestion in our current API views, maybe due to lack of "prior art" because we don't filter anything (yet): https://github.com/ansible-community/ara/blob/master/ara/api/views.py

I guess the "right" way to do it would be via django_filter but I didn't manage to find out a way to do it like that.
Out of curiosity, I almost got it to work in the UI view (even though it's probably not the right place to do it) but the pagination gets in the way: https://github.com/dmsimard/ara/blob/c56719f865233befa3105753f4df5219711920cc/ara/ui/views.py#L76
It expects ``serializer.data`` which is a ``ReturnList`` object, not a ``list(serializer.data)``.

hille721:
> It works for those fields but I haven't been able to figure out the right implementation for the host facts. I wasn't able to find the right way to plumb your suggestion in our current API views, maybe due to lack of "prior art" because we don't filter anything (yet): https://github.com/ansible-community/ara/blob/master/ara/api/views.py

I guess for that we would have to overwrite the `get` method of the viewset. But you won't see those filters than in the browseable API. Have to have a deeper look into that.

> It expects serializer.data which is a ReturnList object, not a list(serializer.data).

You can simply convert it back via:

``` python
from rest_framework.utils import serializer_helpers

response = serializer.data # type(response) = list
response = serializer_helpers.ReturnList(response, serializer=serializer) # type(response) = ReturnList
```

ara/api/models.py
hille721: changed in 60f427f

ara/api/migrations/0009_distincthost.py
```diff
@@ -1,26 +1,8 @@
...
-    updated = []
-    for host in host_model.objects.all():
-        if host.name in updated:
-            continue
-
-        # TODO: Should we sort by another column ?
-        latest_host = host_model.objects.filter(name=host.name).order_by('-updated')[0]
-        distincthost_model.objects.using(db_alias).create(name=host.name, latest_host=latest_host)
-        updated.append(host.name)
-
```
dmsimard: Oops, did you mean to remove that ? We need it.

hille721: sorry, my mistake. Now its there again

dmsimard:
recheck

dmsimard:
I had a bit of time and took a stab at basic tests to begin with, maybe it can help you get started: https://gist.github.com/dmsimard/e82c6a782f87c56cc85898ada56d3a33

The main things I found:
1) Not sure why but the host name argument of the factory doesn't propagate to the subfactory (see TODO)
2) The deletion workflow seems to be tricky to get right. Under normal circumstances, I don't suppose we would delete a distincthost (it's just a relation table) so we would delete a host instead ? Maybe that host isn't the latest one so there is no update necessary to the distincthost table. Maybe it is the latest one so we need to update the latest host to point to the next-latest one. Maybe it is the last one under that name so then we would need to delete the distincthost. Can we make that work ?

dmsimard:
Got a bit more time and wanted to follow up on the deletion workflow from my last comment.

You correctly set ``ReadOnlyModelViewSet`` for the DistinctHost view so we can't create or delete hosts from that endpoint. That's perfect, thanks :)

To demonstrate in other words what I was saying:
- I had 18 hosts named ``localhost`` under ``/api/v1/hosts?name=localhost``
- I queried ``/api/v1/distincthosts?name=localhost``, found the ID of the latest host (say, ``90``) and deleted it via ``/api/v1/hosts``
- The entry from ``/api/v1/distincthosts`` for localhost is gone (but should have been updated to point to the next-latest host)

I suppose we would need to override the deletion method for ``/api/v1/hosts`` to seek the latest host and update it.

dmsimard:
I got deletion to work like this: https://gist.github.com/dmsimard/58c16e3eb9cbb33e34898b0dc2010b59

In essence, before deleting a host, it will check if it is currently the latest host and if it is, it will update it to the next-latest host.

hille721:
> The deletion workflow seems to be tricky to get right. Under normal circumstances, I don't suppose we would delete a distincthost (it's just a relation table) so we would delete a host instead ? Maybe that host isn't the latest one so there is no update necessary to the distincthost table. Maybe it is the latest one so we need to update the latest host to point to the next-latest one. Maybe it is the last one under that name so then we would need to delete the distincthost. Can we make that work ?

-> correct, currently the deletetion was only handeled on DB level, if a related latest host was deleted (`on_delete=models.CASCADE`).  We can easily override this as we do it for creation.


> I suppose we would need to override the deletion method for /api/v1/hosts to seek the latest host and update it.

> I got deletion to work like this: https://gist.github.com/dmsimard/58c16e3eb9cbb33e34898b0dc2010b59
In essence, before deleting a host, it will check if it is currently the latest host and if it is, it will update it to the next-latest host.

-> yes, we have again different possibilities which deletion method we are overriding. You did it now in `views.py`, which works, but we are creating the `DistinctHost` object in the `serializers.py`  (https://github.com/hille721/ara/blob/distinct_hosts/ara/api/serializers.py#L335-L340) thus I would recommend that we are also doing the deletion there for consistency.

hille721:
@dmsimard : as discussed, I've added a `perform_destroy` method in the Host viewset to update the DistinctHost object if necessary. I have also added your unit tests (without modification)

ara/api/views.py
```diff
@@ -95,6 +95,25 @@ class HostViewSet(viewsets.ModelViewSet):
     queryset = models.Host.objects.all()
     filterset_class = filters.HostFilter
 
+    def perform_destroy(self, instance):
```
dmsimard, line 98: Better and safer than my attempt at it I think :+1:

ara/api/tests/tests_distincthost.py
```diff
@@ -0,0 +1,115 @@
+# Copyright (c) 2021 The ARA Records Ansible authors
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+import time
+from rest_framework.test import APITestCase
+
+from ara.api import models, serializers
+from ara.api.tests import factories, utils
+
```
dmsimard, line 9: Oops, copy/pasted imports without checking if they were all used

ara/api/tests/factories.py
```diff
@@ -114,6 +114,12 @@ class Meta:
...
     unreachable = 0
 
+class DistinctHostFactory(DjangoModelFactory):
+    class Meta:
+        model = models.DistinctHost
+
+    name = "hostname"
+    latest = factory.SubFactory(HostFactory, name=name)
 
 class ResultFactory(DjangoModelFactory):
```
dmsimard, line 124:
```suggestion


class DistinctHostFactory(DjangoModelFactory):
    class Meta:
        model = models.DistinctHost

    name = "hostname"
    latest = factory.SubFactory(HostFactory, name=name)


class ResultFactory(DjangoModelFactory):
```

dmsimard:
recheck

dmsimard:
recheck

dmsimard:
It's merged, thanks ! I will finally be able to update the host index PR now :)

ara/api/views.py
dmsimard, line 98: There is a problem with this implementation: this is run when a host is deleted via the API but not when the host is deleted from a cascade reference -- for example, if I delete a playbook, it will delete the hosts referring to it but won't update the latesthost because this part doesn't run.

dmsimard: Tested a fix by adding a similar perform_destroy in the playbook view: https://github.com/ansible-community/ara/pull/308

PR #250, opened 2021-04-07
callback: Truncate play UUIDs to 36 characters

dmsimard:
It was found that ansible-runner suffixes play UUIDs when running in
serial so 34cff6f4-9f8e-6137-3461-000000000005 could become
34cff6f4-9f8e-6137-3461-000000000005_2 and so on.

Truncate them to 36 characters because we don't need the suffix.


dmsimard:
recheck

dmsimard:
recheck

dmsimard:
recheck

PR #245, opened 2021-04-02
Allow user to change theme on the ui directly

guillaumevincent:

PR #251, opened 2021-04-11
UI cleanup

dmsimard:
In a recent PR (https://github.com/ansible-community/ara/pull/245) we moved the theme selection to be client side (in the browser) instead of server side (in the django settings).
This cleans up and tidies things up a bit.


dmsimard:
@guillaumevincent o/ FYI if you would like to review

It's up on https://dev.demo.recordsansible.org/ if you would like to look at it.

PR #252, opened 2021-04-13
Use link in pagination instead of button

guillaumevincent:
image

instead of 

image

### DIFF
--- a/ara/ui/templates/partials/pagination.html
+++ b/ara/ui/templates/partials/pagination.html
@@ -18,10 +18,8 @@
         <a class="page-link" href="#" aria-label="Go to previous page"><span aria-hidden="true">&laquo;</span></a>
       </li>
     {% endif %}
-    <li class="disabled">
-      <div>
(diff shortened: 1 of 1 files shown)


dmsimard:
@guillaumevincent it was a link before https://github.com/ansible-community/ara/commit/be018849847a3baad687943b3a66da5fd6c42f48 but the information in the middle is not a link, though. I was looking for a way to "split" the pagination up a bit because there are otherwise no division (lines) between any of the actual links and it didn't feel optimal, particularly in light theme.

I realize it shouldn't be a "fake" button but I didn't find another styling class that worked and aligned well. Open to suggestions :)

guillaumevincent:
my suggestion is to avoid using a button and use a simple pagination with "first | < | current page | > | last"

dmsimard:
So the ``1-6 of X`` would be outside the pagination button group ? Maybe above or below ?

dmsimard:
I don't feel strong enough in either way, let's go with consistency and keep the link class for now.

guillaumevincent:
yes per page element on the left and navigation on the right

image

dmsimard:
@guillaumevincent good idea, we can do it in the future and go with the current implementation for now.

PR #258, opened 2021-04-14
setup.cfg: pin tzlocal to <3.0

dmsimard:
tzlocal has a pre-release for 3.0 dating back to 2020 and it breaks when
installing ara with "pip install --pre ara[server]". We can revisit this
if they do end up doing an actual release.


dmsimard:
Tested locally, merging manually due to Zuul job reliability issues.

Issue #263, opened 2021-04-15
sqlite3.OperationalError: no such table: playbooks


strus38:
## What is the issue ?
Reading the install doc 
```
# Install Ansible and ARA (with API server dependencies) for the current user
python3 -m pip install --user ansible "ara[server]"

# Configure Ansible to use the ARA callback plugin
export ANSIBLE_CALLBACK_PLUGINS="$(python3 -m ara.setup.callback_plugins)"

# Run an Ansible playbook
ansible-playbook playbook.yaml
```

and then running it, fails with:
```
$ ara playbook list
2021-04-15 20:01:23,224 ERROR django.request: Internal Server Error: /api/v1/playbooks
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/sqlite3/base.py", line 383, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.OperationalError: no such table: playbooks
```

## What should be happening ?
It should simply list nothing or the executed playbooks as stated in the documentation.

dmsimard:
Hi @strus38 and thanks for the issue !

The callback should take care of running the SQL migrations automatically by default and you should actually see the SQL migration messages. If you try to run commands without first recording a playbook or running ``ara-manage migrate`` manually first, you might bump into that error.

Could you run the command ``python3 -m ara.setup.callback_plugins`` manually and see if that returns the path to the callback plugin ?

strus38:
Hi
After running a playbook which does something, it works better indeed. However I think the tool could be enhanced to detect an empty sqlite3 file and display a more friendly error like: 'Empty database, please run a playbook to initialize the DB tables'.
Thanks!
Working like a charm now.

dmsimard:
> However I think the tool could be enhanced to detect an empty sqlite3 file and display a more friendly error like: 'Empty database, please run a playbook to initialize the DB tables'.

Yes, we could probably do something like that. I'll keep the issue opened so we can take care of it. Thanks for the feedback !

dmsimard:
@strus38 FYI I've just created a troubleshooting documentation page where I include this issue: https://ara.readthedocs.io/en/latest/troubleshooting.html

We can figure out how to make the exception friendlier later.

PR #266, opened 2021-04-19
#263: Update readme with migrate database

Ymil:


dmsimard:
Hey @Ymil and thanks for the PR.

I'll close this since the getting started guide purposefully doesn't contain a step for running the SQL migrations because it is not necessary: they are run automatically by the callback plugin, or the container image when going that route. If you bump into a missing database, it's because you would not have run a playbook like the guide suggests.

You can see the callback running migrations automatically in this [gif of the README](https://github.com/ansible-community/ara/blob/master/doc/source/_static/ara-quickstart-default.gif).

However, I think we could do a better job from a documentation (or exception handling) perspective as I wrote in another issue: https://github.com/ansible-community/ara/issues/263

PR #268, opened 2021-04-20
Add caching for tasks in callback plugin

jhampson-dbre:
When running a playbook with free strategy, multiple hosts could be running different playbook tasks at the same time. This caused ARA to associate results to the wrong task id, resuting in task names not matching the task result in the ARA web interface.

Now, the task is cached by it's uuid. The task result includes the task uuid, so this is used to ensure the correct task id is associate with a particular result.

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -372,8 +372,7 @@ class CallbackModule(CallbackBase):
         task_file = self._get_or_create_file(path)
 
         # Get task
-        self.task = self._get_or_create_task(
-            task, task_file["id"], lineno, handler)
+        self.task = self._get_or_create_task(task, task_file["id"], lineno, handler)
 
         return self.task
 
@@ -491,8 +490,7 @@ class CallbackModule(CallbackBase):
         task_uuid = str(task._uuid)[:36]
         if task_uuid not in self.task_cache:
             if None in (file_id, lineno, handler):
-                raise ValueError(
-                    "file_id, lineno, and handler are required to create a task")
+                raise ValueError("file_id, lineno, and handler are required to create a task")
 
             self.log.debug("Task not in cache, getting or creating: %s" % task)
             self.task_cache[task_uuid] = self.client.post(
@@ -506,8 +504,7 @@ class CallbackModule(CallbackBase):
                 tags=task.tags,
                 lineno=lineno,
                 handler=handler,
-                started=datetime.datetime.now(
-                    datetime.timezone.utc).isoformat(),
+                started=datetime.datetime.now(datetime.timezone.utc).isoformat(),
             )
 
         return self.task_cache[task_uuid]
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -205,6 +205,7 @@ class CallbackModule(CallbackBase):
         self.stats = None
         self.file_cache = {}
         self.host_cache = {}
+        self.task_cache = {}
 
     def set_options(self, task_keys=None, var_options=None, direct=None):
         super(CallbackModule, self).set_options(task_keys=task_keys, var_options=var_options, direct=direct)
@@ -370,19 +371,9 @@ class CallbackModule(CallbackBase):
         # Get task file
         task_file = self._get_or_create_file(path)
 
-        self.task = self.client.post(
-            "/api/v1/tasks",
-            name=task.get_name(),
-            status="running",
-            action=task.action,
-            play=self.play["id"],
-            playbook=self.playbook["id"],
-            file=task_file["id"],
-            tags=task.tags,
-            lineno=lineno,
-            handler=handler,
-            started=datetime.datetime.now(datetime.timezone.utc).isoformat(),
-        )
+        # Get task
+        self.task = self._get_or_create_task(
+            task, task_file["id"], lineno, handler)
 
         return self.task
 
@@ -495,6 +486,32 @@ class CallbackModule(CallbackBase):
             self.host_cache[host] = self.client.post("/api/v1/hosts", name=host, playbook=self.playbook["id"])
         return self.host_cache[host]
 
+    def _get_or_create_task(self, task, file_id=None, lineno=None, handler=None):
+        # Note: The get_or_create is handled through the serializer of the API server.
+        task_uuid = str(task._uuid)[:36]
+        if task_uuid not in self.task_cache:
+            if None in (file_id, lineno, handler):
+                raise ValueError(
+                    "file_id, lineno, and handler are required to create a task")
+
+            self.log.debug("Task not in cache, getting or creating: %s" % task)
+            self.task_cache[task_uuid] = self.client.post(
+                "/api/v1/tasks",
+                name=task.get_name(),
+                status="running",
+                action=task.action,
+                play=self.play["id"],
+                playbook=self.playbook["id"],
+                file=file_id,
+                tags=task.tags,
+                lineno=lineno,
+                handler=handler,
+                started=datetime.datetime.now(
+                    datetime.timezone.utc).isoformat(),
+            )
+
+        return self.task_cache[task_uuid]
+
     def _load_result(self, result, status, **kwargs):
         """
         This method is called when an individual task instance on a single
@@ -506,6 +523,9 @@ class CallbackModule(CallbackBase):
 
         # Retrieve the host so we can associate the result to the host id
         host = self._get_or_create_host(hostname)
+
+        # Retrieve the task so we can associate the result to the task id
+        task = self._get_or_create_task(result._task)
 
         results = strip_internal_keys(module_response_deepcopy(result._result))
 
@@ -529,19 +549,19 @@ class CallbackModule(CallbackBase):
         self.result = self.client.post(
             "/api/v1/results",
             playbook=self.playbook["id"],
-            task=self.task["id"],
+            task=task["id"],
             host=host["id"],
-            play=self.task["play"],
+            play=task["play"],
             content=results,
             status=status,
-            started=self.result_started[hostname] if hostname in self.result_started else self.task["started"],
+            started=self.result_started[hostname] if hostname in self.result_started else task["started"],
             ended=self.result_ended[hostname],
             changed=result._result.get("changed", False),
             # Note: ignore_errors might be None instead of a boolean
             ignore_errors=kwargs.get("ignore_errors", False) or False,
         )
 
-        if self.task["action"] in ["setup", "gather_facts"] and "ansible_facts" in results:
+        if task["action"] in ["setup", "gather_facts"] and "ansible_facts" in results:
             self.client.patch("/api/v1/hosts/%s" % host["id"], facts=results["ansible_facts"])
 
     def _load_stats(self, stats):


dmsimard:
Hey @jhampson-dbre and thanks for the PR!

There is nothing shocking to me in there and it looks good to me at first glance but I am curious if there are any performance or memory usage implications in the change. I'll try and find time to check it out sometime soon a bit like I did when I [benchmarked the callback overhead](https://ara.recordsansible.org/blog/2020/11/01/benchmarking-ansible-and-ara-for-fun-and-science/) since we already have the numbers and I'll get back to you.

jhampson-dbre:
I don't really understand `_end_task`

https://github.com/ansible-community/ara/blob/2867f2dad2121bb8f5132f88cc93b8d36b35f8db/ara/plugins/callback/ara_default.py#L402-L410

Since you would have the `_end_task` being called multiple times for a particular task on different hosts (with strategy free), I'm not sure the task endtime/status would be accurately recorded.

dmsimard:
The purpose of end task is to flip the task status to completed and put an end date so we can calculate the duration of the task across all hosts who ran that task.

In linear, it is used in places where there is a new task starting so we know that the previous task has ended. So you're right, I guess in free we can't rely on that anymore :)

We need to be able to tell that there are no more expected results for that task (or if we know it is the last result for a task) in order to set the status and the end date.

jhampson-dbre:
Would it make sense to have end time be a property of result instead of task? Then duration would be task start time - result time.  

Would also be worth checking of the result object from Ansible contains time or duration properties.

dmsimard:
btw: I am not a unit test person but to me this screams "integration test" -- we didn't really know that this was broken because we don't have any playbooks running in free, much less validating that their contents is appropriate.

There's a tox environment that runs [integration tests](https://github.com/ansible-community/ara/blob/master/tests/basic.yaml) with the callback enabled: ``tox -e ansible-integration``. We could add a playbook in there that runs with the free strategy and asserts that what the API returns makes sense like in the [lookup tests](https://github.com/ansible-community/ara/blob/master/tests/integration/lookups.yaml).

dmsimard:
> Would it make sense to have end time be a property of result instead of task? Then duration would be task start time - result time.

Funnily enough, that's actually more or less how it was done before we moved the duration to the database instead of "computing" it every time: https://github.com/ansible-community/ara/commit/1f99014dbc5c20e92909024c949b7023831d3299

If I recall properly, the problem we had is that as a property, sorting would need to happen outside the database whereas sorting the field in-database is super fast and "free". Sorting things by duration is pretty cool.

> Would also be worth checking of the result object from Ansible contains time or duration properties.

I haven't looked in what is probably a few years so it's probably worth revisiting if the data is there.

The way the duration is currently captured includes the overhead that recording with ara implies and I think that is an honest representation because... that's the time it took. Data from Ansible would allow us to do simple math to tell how much time ara really added to a task and that would be interesting :)

jhampson-dbre:
After looking at it again, it looks like the start and end time is already capture per host and stored with the result. 
https://github.com/ansible-community/ara/blob/2867f2dad2121bb8f5132f88cc93b8d36b35f8db/ara/plugins/callback/ara_default.py#L554-L555

I think this is what is displayed in the ARA web ui, so I'm not sure most people would see the task times (as opposed to the result times) without querying the api or doing custom reporting from the DB. 

In either case, the task timing should continue to work the same as before for linear strategy, and for free strategy I'm not sure the total task duration is of much use since you could have a long gap where the task isn't even running on any host, but the task timer is still running because some hosts haven't even reached the task yet.

dmsimard:
> I think this is what is displayed in the ARA web ui, so I'm not sure most people would see the task times (as opposed to the result times) without querying the api or doing custom reporting from the DB.

Yes, you are correct that the task duration is not implemented in the UI (yet) but it's already used in the CLI, for example:
```bash
> ara task list --order=-duration
+--------+-----------+---------+----------------------------+-------------------------------------------------+----------+-----------------------------+-----------------+
|     id | status    | results | action                     | name                                            | playbook | started                     | duration        |
+--------+-----------+---------+----------------------------+-------------------------------------------------+----------+-----------------------------+-----------------+
| 304695 | completed |       1 | shell                      | os_tempest : Execute tempest tests              |     1661 | 2020-12-21T17:17:49.501000Z | 00:13:46.034681 |
| 257620 | completed |       1 | community.mysql.mysql_user | galera_server : Create galera users             |     1565 | 2020-12-21T15:45:14.185000Z | 00:10:34.464058 |
| 299621 | completed |       1 | command                    | Waiting for rabbitmq to start                   |     1683 | 2020-12-21T16:44:47.272000Z | 00:10:18.162493 |
|  19434 | completed |       1 | shell                      | os_tempest : Execute tempest tests              |       34 | 2020-01-31T17:43:02.943000Z | 00:08:21.191461 |
# ...
```

or:
```bash
> ara task metrics --aggregate name --path openstack-ansible --limit 2000 --sort-column=duration_total --sort-descending
+---------------------------------------------------------------------------------------+-------+---------+----------------+----------------+
| aggregate                                                                             | count | results | duration_total | duration_avg   |
+---------------------------------------------------------------------------------------+-------+---------+----------------+----------------+
| os_tempest : Execute tempest tests                                                    |     7 |       7 | 0:40:08.829470 | 0:05:44.118496 |
| os_neutron : Stop services                                                            |    10 |      10 | 0:05:32.294844 | 0:00:33.229484 |
| Execute service action                                                                |    11 |      11 | 0:02:13.893420 | 0:00:12.172129 |
| os_tempest : Install distro packages                                                  |     5 |       5 | 0:01:49.398349 | 0:00:21.879670 |
| os_neutron : Install neutron role packages                                            |     7 |       7 | 0:01:24.469024 | 0:00:12.067003 |
# ...
```

It's particularly useful because you can't infer the duration of a task based on a single result since it kind of goes like this:

```
                                     task
<------------------------------------------------------------------------>

    host1
<---------->
         host2
    <------------>
                                              host3
               <--------------------------------------------------------->
# ...
```

> In either case, the task timing should continue to work the same as before for linear strategy

Right.

> for free strategy I'm not sure the total task duration is of much use since you could have a long gap where the task isn't even running on any host, but the task timer is still running because some hosts haven't even reached the task yet.

Yeah, I suppose task durations aren't particularly useful or relevant when using the free strategy. I don't have any great ideas to tackle that right now and I think I can live with them being somewhat bogus for now, especially considering it wasn't working at all before, it's an improvement over what we had.

I guess we still need to figure out how to tell that the result we are getting is the last result for the task so we can mark the task as completed. Worst case scenario, we could end up posting results to a task that was already marked as completed which isn't the end of the world and the API doesn't verify that but it would be nice to find something more reliable.

Otherwise, it makes me think that, to my knowledge, ara doesn't pick up the strategy that a playbook is running in. We should probably do that if it's available from the callback interface. I created an issue for this here: https://github.com/ansible-community/ara/issues/269

jhampson-dbre:
I was looking around at other callback plugins to see how they handled free. Most don't seem to account for it. 🤷‍♂️

One I found that does do some special things for free strategy is counter_enabled . It get's the play variables during v2_playbook_on_play_start from `play.get_variable_manager().get_vars(play)`, which includes the `strategy` play variable :

https://github.com/ansible/ansible/blob/473608521f9dbc2a2104c26785554a5f6dc6d58b/lib/ansible/plugins/callback/counter_enabled.py#L53-L60

```python
    def _all_vars(self, host=None, task=None):
        # host and task need to be specified in case 'magic variables' (host vars, group vars, etc)
        # need to be loaded as well
        return self._play.get_variable_manager().get_vars(
            play=self._play,
            host=host,
            task=task
        )
```

For task completion tracking, maybe you could track when the number of results for a task >= the number of hosts remaining in the play (as indicated by the current value of `ansible_play_hosts`, assuming you can get the current value as the playbook is running to account for hosts that have failed). 

Alternatively, consider a task open until the play ends (when play strategy is free). Continue to update the task end time as results come in (which is already happening), and then when the play ends, mark all tasks as complete. It wouldn't be 100% realtime update that the task was complete, but it would be more accurate than showing it is complete when there are still hosts that have yet to run.

dmsimard:
o/ I haven't forgotten about this but haven't had the opportunity to come back to it yet. Another user reported issues when running in free and It's still on my radar.

I would like to have a fix merged in time for the upcoming release.

dmsimard:
o/ @jhampson-dbre

I scratched my head a bit but I believe I've figured out an integration test for this -- it works even if it is quite tricky to say the least :)

The playbook is in this gist: https://gist.github.com/dmsimard/d249404addeb889317bd835cc66a7b0f

Here's what it looks like without your patch:
```
> ara task list --playbook 7
+-----+-----------+---------+--------------+----------------------------------------------------+----------+----------------------------------+-----------------+
|  id | status    | results | action       | name                                               | playbook | started                          | duration        |
+-----+-----------+---------+--------------+----------------------------------------------------+----------+----------------------------------+-----------------+
| 167 | completed |       1 | assert       | Assert the number of tasks                         |        7 | 2021-07-20T16:51:35.861384-04:00 | 00:00:00.110695 |
| 166 | completed |       1 | set_fact     | Search for the free tasks                          |        7 | 2021-07-20T16:51:35.664762-04:00 | 00:00:00.180232 |
| 165 | completed |       1 | set_fact     | Search for the free play                           |        7 | 2021-07-20T16:51:35.476356-04:00 | 00:00:00.122795 |
| 164 | completed |       1 | ara_playbook | Retrieve the current playbook so we can get the ID |        7 | 2021-07-20T16:51:35.269274-04:00 | 00:00:00.150531 |
| 163 | completed |       3 | command      | Third task                                         |        7 | 2021-07-20T16:51:32.883172-04:00 | 00:00:02.279132 |
| 162 | completed |       4 | command      | Third task                                         |        7 | 2021-07-20T16:51:32.061082-04:00 | 00:00:00.763367 |
| 161 | completed |       1 | command      | Third task                                         |        7 | 2021-07-20T16:51:31.319372-04:00 | 00:00:00.679806 |
| 160 | completed |       1 | command      | Third task                                         |        7 | 2021-07-20T16:51:31.172727-04:00 | 00:00:00.115788 |
| 159 | completed |       0 | command      | Third task                                         |        7 | 2021-07-20T16:51:31.071953-04:00 | 00:00:00.071776 |
| 158 | completed |       0 | command      | Third task                                         |        7 | 2021-07-20T16:51:31.010088-04:00 | 00:00:00.030476 |
| 157 | completed |       7 | command      | Second task                                        |        7 | 2021-07-20T16:51:30.682573-04:00 | 00:00:00.268987 |
| 156 | completed |       0 | command      | Third task                                         |        7 | 2021-07-20T16:51:30.585860-04:00 | 00:00:00.034933 |
| 155 | completed |       0 | command      | Second task                                        |        7 | 2021-07-20T16:51:30.480637-04:00 | 00:00:00.073550 |
| 154 | completed |       0 | command      | Third task                                         |        7 | 2021-07-20T16:51:30.359859-04:00 | 00:00:00.090599 |
| 153 | completed |       0 | command      | Third task                                         |        7 | 2021-07-20T16:51:30.268156-04:00 | 00:00:00.066535 |
| 152 | completed |       0 | command      | Third task                                         |        7 | 2021-07-20T16:51:30.214762-04:00 | 00:00:00.030209 |
| 151 | completed |       6 | command      | Second task                                        |        7 | 2021-07-20T16:51:29.803359-04:00 | 00:00:00.350130 |
| 150 | completed |       0 | command      | Second task                                        |        7 | 2021-07-20T16:51:29.603027-04:00 | 00:00:00.170964 |
| 149 | completed |       0 | command      | Second task                                        |        7 | 2021-07-20T16:51:29.378852-04:00 | 00:00:00.203432 |
| 148 | completed |       0 | command      | Second task                                        |        7 | 2021-07-20T16:51:29.253622-04:00 | 00:00:00.103246 |
| 147 | completed |       0 | command      | Second task                                        |        7 | 2021-07-20T16:51:29.198107-04:00 | 00:00:00.032583 |
| 146 | completed |       0 | command      | Second task                                        |        7 | 2021-07-20T16:51:29.117680-04:00 | 00:00:00.020500 |
| 145 | completed |       0 | command      | Second task                                        |        7 | 2021-07-20T16:51:29.027797-04:00 | 00:00:00.067927 |
| 144 | completed |       0 | command      | Second task                                        |        7 | 2021-07-20T16:51:28.966808-04:00 | 00:00:00.028587 |
| 143 | completed |       8 | command      | First task                                         |        7 | 2021-07-20T16:51:28.299094-04:00 | 00:00:00.606042 |
| 142 | completed |       0 | command      | First task                                         |        7 | 2021-07-20T16:51:27.567337-04:00 | 00:00:00.705251 |
| 141 | completed |       0 | command      | First task                                         |        7 | 2021-07-20T16:51:27.343222-04:00 | 00:00:00.202841 |
| 140 | completed |       0 | command      | First task                                         |        7 | 2021-07-20T16:51:27.249192-04:00 | 00:00:00.071773 |
| 139 | completed |       0 | command      | First task                                         |        7 | 2021-07-20T16:51:27.134565-04:00 | 00:00:00.090651 |
| 138 | completed |       0 | command      | First task                                         |        7 | 2021-07-20T16:51:27.045380-04:00 | 00:00:00.023243 |
| 137 | completed |       0 | command      | First task                                         |        7 | 2021-07-20T16:51:26.943054-04:00 | 00:00:00.072185 |
| 136 | completed |       0 | command      | First task                                         |        7 | 2021-07-20T16:51:26.889938-04:00 | 00:00:00.027507 |
| 135 | completed |       0 | command      | First task                                         |        7 | 2021-07-20T16:51:26.809999-04:00 | 00:00:00.019583 |
| 134 | completed |       0 | command      | First task                                         |        7 | 2021-07-20T16:51:26.728270-04:00 | 00:00:00.063084 |
| 133 | completed |       1 | add_host     | Add a host to the inventory                        |        7 | 2021-07-20T16:51:26.449592-04:00 | 00:00:00.153730 |
+-----+-----------+---------+--------------+----------------------------------------------------+----------+----------------------------------+-----------------+
```

The above fails on the first assert that checks that we should have three tasks -- we somehow ended up with 30 tasks !

And then with this PR:
```
> ara task list --playbook 6
+-----+-----------+---------+--------------+----------------------------------------------------+----------+----------------------------------+-----------------+
|  id | status    | results | action       | name                                               | playbook | started                          | duration        |
+-----+-----------+---------+--------------+----------------------------------------------------+----------+----------------------------------+-----------------+
| 132 | completed |       1 | assert       | Assert the status of the tasks                     |        6 | 2021-07-20T16:50:06.315535-04:00 | 00:00:00.096365 |
| 131 | completed |       1 | assert       | Assert the number of tasks                         |        6 | 2021-07-20T16:50:06.182180-04:00 | 00:00:00.074052 |
| 130 | completed |       1 | set_fact     | Search for the free tasks                          |        6 | 2021-07-20T16:50:05.958574-04:00 | 00:00:00.165040 |
| 129 | completed |       1 | set_fact     | Search for the free play                           |        6 | 2021-07-20T16:50:05.771771-04:00 | 00:00:00.163036 |
| 128 | completed |       1 | ara_playbook | Retrieve the current playbook so we can get the ID |        6 | 2021-07-20T16:50:05.588381-04:00 | 00:00:00.164350 |
| 127 | completed |      10 | command      | Third task                                         |        6 | 2021-07-20T16:49:59.279908-04:00 | 00:00:06.199003 |
| 126 | completed |      10 | command      | Second task                                        |        6 | 2021-07-20T16:49:57.604225-04:00 | 00:00:03.293001 |
| 125 | completed |      10 | command      | First task                                         |        6 | 2021-07-20T16:49:54.600208-04:00 | 00:00:02.981818 |
| 124 | completed |       1 | add_host     | Add a host to the inventory                        |        6 | 2021-07-20T16:49:54.285418-04:00 | 00:00:00.168842 |
+-----+-----------+---------+--------------+----------------------------------------------------+----------+----------------------------------+-----------------+
```

That's much better.

In light of these results, any objections on whether we should land this as it is or not ? Have you been running it in production and how has it been working for you ?
I can open a PR of my own to add the integration tests.

It doesn't address the task completion flip (due to end_task being at the beginning of a task) but we can take care of that in another issue/PR.

jhampson-dbre:
We've been running it with the proposed changes for a while. Without the changes, it's basically unusable with free strategy. I think it's good to ship.

dmsimard:
> We've been running it with the proposed changes for a while. Without the changes, it's basically unusable with free strategy. I think it's good to ship.

In addition to the tests that now work, I've also received additional feedback that it works for other users.

Thanks ! It will be part of the 1.5.7 release.

PR #274, opened 2021-05-05
ara_playbook/ara_record: set BYPASS_HOST_LOOP to run once

dmsimard:
These action plugins interact with the API to get a playbook or add a
key/value record. When run against an inventory of 100 hosts, these
shouldn't be running 100 times as they only need to run once to retrieve
or set the data.

### DIFF
--- a/ara/plugins/action/ara_playbook.py
+++ b/ara/plugins/action/ara_playbook.py
@@ -67,6 +67,9 @@ class ActionModule(ActionBase):
     """ Retrieves either a specific playbook from ARA or the one currently running """
 
     TRANSFERS_FILES = False
+    # BYPASS_HOST_LOOP functions like a forced "run_once" on a task
+    # We don't need to run this module for every host.
+    BYPASS_HOST_LOOP = True
     VALID_ARGS = frozenset(("playbook_id"))
 
     def __init__(self, *args, **kwargs):
--- a/ara/plugins/action/ara_record.py
+++ b/ara/plugins/action/ara_record.py
@@ -135,6 +135,9 @@ class ActionModule(ActionBase):
     """ Record persistent data as key/value pairs in ARA """
 
     TRANSFERS_FILES = False
+    # BYPASS_HOST_LOOP functions like a forced "run_once" on a task
+    # We don't need to run this module for every host.
+    BYPASS_HOST_LOOP = True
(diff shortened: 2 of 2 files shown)


flowerysong:
-1

ara_playbook has a stronger case for using this, but even there I could see someone wanting to retrieve a different playbook based on the current host (e.g. storing the playbook ID that generated a machine image in a file, then using the contents of that file as `playbook_id`).

ara_record can currently be used for storing per-host records, which this change would prevent.

dmsimard:
Hey @flowerysong, thanks for the input and you're right, I agree.

This originates from a user asking about why the modules were running N times for N hosts and perhaps the better approach is instead to add a suggestion/recommendation to set ``run_once`` on the task unless there is a use case like the ones you mentioned.

dmsimard:
Putting this in draft, I have no intention of merging it but keeping it opened as a reminder for now.

dmsimard:
We're four years later and I would like to revisit this as part of adding another action module: [ara_label](https://github.com/ansible-community/ara/issues/596).

I can see and agree about the use case for running ara_record, ara_playbook and ara_label for every host to get dynamically templated values, etc.

I pondered on making [run_once](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_strategies.html#running-on-a-single-machine-with-run-once) the default and letting users run it across every host if they really mean to.

If I ran a playbook against a hundred hosts and had a task to set a playbook label like this:
```
- name: Add labels to this playbook (the one that is running)
  ara_label:
    state: present
    labels:
      - running
```

It would set the label running a hundred times. That's not great.

If someone wanted to include host/templated values, they could set run_once to false:
```
- name: Add labels to this playbook (the one that is running)
  ara_label:
    state: present
    run_once: false
    labels:
      - running
      - "os:{{ ansible_distribution }}-{{ ansible_distribution_version }}"
```

This is anti-pattern to the ansible docs about using ``run_once: true`` so it would warrant being documented.

Or maybe we could leave things the way they are and let users specify ``run_once: true`` unless they mean to iterate over the hosts. This would benefit from better documentation as well.

From personal usage and feedback from users, it is a bit "foot-gun" if you forget or don't realize what it does.
Either way we go we need to make the behavior a bit more explicit.

dmsimard:
Closing this PR, it has been migrated to codeberg: https://codeberg.org/ansible-community/ara/pulls/274

PR #279, opened 2021-06-10
Unpin SQLAlchemy version

javierpena:
Previously, the SQLAlchemy version was pinned to < 1.3.0 due to
an issue with flask-sqlalchemy. This was fixed long ago, and the
current requirements.txt causes an issue at least with Python 2,
since it installs incompatible combinations of SQLAlchemy and alembic.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,9 +2,7 @@ pbr!=2.1.0,>=2.0.0 # Apache-2.0
 
 jinja2>=2.9
 ansible>=2.6.0
-# Pin SQLAlchemy until new release of flask-sqlalchemy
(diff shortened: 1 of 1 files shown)


javierpena:
I've found a workaround that doesn't require reviving an old branch.

dmsimard:
@javierpena for the sake of posterity could you share the workaround ?

wisdom_of_the_ancients

javierpena:
I'm working around it with https://review.rdoproject.org/r/c/rdo-infra/weirdo/+/34061 . Since the issue is not properly fixed by the pip < 21 resolver, I just "pip install alembic<1.5.0 SQLAlchemy<1.3.0" before installing ara.

Issue #256, opened 2021-04-14
UI: Display task tags


dmsimard:
## What is the idea ?

ARA saves task tags but we're not showing them anywhere in the UI right now.

In 0.x it used to be a tooltip over the task name: https://github.com/ansible-community/ara/blob/stable/0.x/ara/templates/ajax/task_name.html#L1-L13

It should also be included in the task result details.

PR #281, opened 2021-06-26
Add tags in task results details. Add tags in as tooltip in task results

JayV3e:
This fix issue #256.

In task results, I added a tag button with the number of tags.
On mouse over, there is a tooltip that shows the tags.

Let me know if I should stick to the previous pattern of mouse over the task name instead of an additional span.

Screen Shot 2021-06-26 at 5 12 08 PM

In task result, I added the Tags if there is any. Otherwise, the Tags: won't show up.

With tags : 
Screen Shot 2021-06-26 at 5 12 21 PM

Without tags : 
Screen Shot 2021-06-26 at 5 12 31 PM

I added the <script> and <style> directly in the playbook_results.html template.
I'm not sure if that's a good practice or not. 

Waiting for feedback on the tag span, where to put the <script> and <style> or any additional feedback. 

This is my first pull request on an open source project, let me know if I forgot anything relevant.


ara/ui/templates/partials/playbook_results.html
```diff
@@ -149,6 +149,11 @@ <h4>
                     {% else %}
                       {{ result.task.name }}
                     {% endif %}
+                    {% if result.task.tags %}
+                    <span class="badge badge-pill badge-warning" class="btn btn-secondary" data-toggle="tooltip"  data-placement="right" data-html="true" title="{% include 'partials/tags_list.html' with tags=result.task.tags %}">
```
dmsimard, line 153: I'm OK with this approach but I would tend to use another color here because it overlaps with what we use to highlight the changed status:
Screenshot from 2021-06-26 21-05-09

Looking at the [docs](https://getbootstrap.com/docs/4.0/components/badge/#contextual-variations), maybe ``badge-secondary`` instead of ``badge-warning`` ?

dmsimard:
It looks good to me although I wonder if it's the right place or column for the badge pill because tasks have variable length names so the badge can display a bit all over the place. I wonder what it would look like in the status column ?

We would also need to add the tags in the same place for the host results table: https://github.com/ansible-community/ara/blob/master/ara/ui/templates/partials/host_results.html#L160

dmsimard:
Also: ignore the failing container job, it's unrelated to your patch and I need to fix it.

JayV3e:
I changed color and column.

By default, it looks awful : 
Screen Shot 2021-06-27 at 4 34 09 PM

We could either add the label with 0 tags while removing the empty tooltip : 
Screen Shot 2021-06-27 at 4 34 34 PM

Or simply add "invisible" as a class for the span : 
Screen Shot 2021-06-27 at 4 35 16 PM

I also added the same label to the host result : 
Screen Shot 2021-06-27 at 4 43 15 PM


Initially I thought invisible would've been the best solution but I kinda like the 0 tags as well.

Let me know whichever you prefer.

flowerysong:
While this should definitely be included in the detailed view, I don't really feel this is important enough information to surface it so prominently.

JayV3e:
I agree. My example only has 3 tags and it already takes quite some space. I don't know how many tags people use on average but this could easily get out of hands.

Here's 3 suggestions : 
- Make a list of the tags

Ex : 
Tags: tag1,tag2,tag3,tag4

- Keep it as it is but make it collapsed by default and expendable.

- Use the same pattern as in the playbook result: Have a label to show to number of tags and have the list appear on mouse over.

My vote goes for the last one.

Let me know if you have another suggestion.

dmsimard:
> While this should definitely be included in the detailed view, I don't really feel this is important enough information to surface it so prominently.

Thanks for the feedback @flowerysong !
I agree to some extent which is in part why tags had not yet made their way to the results table yet :)

The task tags are included by default "for free" in the API response, we're just not displaying them anywhere right now. I figure it would be useful to at least do something with them.

In ara 0.x, it used to be a tooltip over the task name, like this: https://github.com/ansible-community/ara/blob/stable/0.x/ara/templates/ajax/task_name.html
Screenshot from 2021-06-30 15-44-29

That implementation wasn't perfect either because you only knew if the task had tags at all if you put your mouse over the task name so it wasn't obvious from a UX perspective which tasks had tags or not.

Maybe we could go with that approach instead ? There is no debate to be had about the detailed view, we all agree it would be a nice addition.

dmsimard:
Hi @JayV3e o/

Sorry about the delay -- I've taken a bit of time to think about this and asked different users what they thought.

Given the choice, not everyone prefers the same one (UI/UX is hard!) but there's a small majority that would go with the third option you suggested:
image

If that's okay with you, we'll go with that for now and I'm happy to revisit later if we find a better implementation.

dmsimard:
This isn't a blocker for releasing 1.5.7 which I would like to tag, it can go in the next release.

Feel free to let me know when you see my comments @JayV3e o/

ara/ui/templates/partials/host_results.html
```diff
@@ -145,6 +145,13 @@
                     {% else %}
                       {% include "partials/result_status_icon.html" with status=result.status %}
                     {% endif %}
+                    {% if result.task.tags|length == 0 %}
```
dmsimard, line 148: nitpick, would rather have the version with the space like the one you wrote below for consistency

```suggestion
                    {% if result.task.tags | length == 0 %}
```

ara/ui/templates/partials/playbook_results.html
```diff
@@ -127,6 +127,13 @@ <h4>
                     {% else %}
                       {% include "partials/result_status_icon.html" with status=result.status %}
                     {% endif %}
+                    {% if result.task.tags|length == 0 %}
```
dmsimard, line 130: ditto
```suggestion
                    {% if result.task.tags | length == 0 %}
```

ara/ui/templates/partials/tags_list.html
```diff
@@ -0,0 +1,5 @@
+<ul>
+    {% for tag in tags %}
+    <li> {{tag}} </li>
```
dmsimard, line 3:
```suggestion
    <li> {{ tag }} </li>
```

JayV3e:
Hello,

I applied all suggestions and changed the tags formatting in task results to use bullet points.

Hope I did it in time for the release !

dmsimard:
The patch would have been fine but the CI is now reporting an issue with the PR when running ``ara-manage generate`` to generate a static version of the report:
```
[ara] Using settings file: /home/zuul/.ara-tests/server/settings.yaml
[ara] Generating static files for 8 playbooks at /home/zuul/.ara-tests/server/static...
Traceback (most recent call last):
  File "/home/zuul/.ara-tests/virtualenv/lib64/python3.9/site-packages/django/template/smartif.py", line 175, in translate_token
    op = OPERATORS[token]
KeyError: '|'

During handling of the above exception, another exception occurred:
# ...
  File "/home/zuul/.ara-tests/virtualenv/lib64/python3.9/site-packages/django/template/base.py", line 662, in __init__
    raise TemplateSyntaxError("Could not parse the remainder: '%s' "
django.template.exceptions.TemplateSyntaxError: Could not parse the remainder: '|' from '|'
```

That's odd because it was green until the last commit:
Screenshot from 2021-08-02 14-28-26

I will need to investigate after the release. Sorry !

ara/ui/templates/partials/host_results.html
```diff
@@ -145,6 +145,13 @@
                     {% else %}
                       {% include "partials/result_status_icon.html" with status=result.status %}
                     {% endif %}
+                    {% if result.task.tags | length == 0 %}
+                      <span class="badge badge-pill badge-secondary invisible" class="btn btn-secondary">
+                    {% else %}
+                      <span class="badge badge-pill badge-secondary" class="btn btn-secondary" data-toggle="tooltip"  data-placement="right" data-html="true" title="{% include 'partials/tags_list.html' with tags=result.task.tags %}">
+                    {% endif %}
+                      {{ result.task.tags | length }} tags
+                      </span>
```
dmsimard, line 154: I was able to reproduce the issue I mentioned in my [last comment](https://github.com/ansible-community/ara/pull/281#issuecomment-891240781) locally and it's indeed tripping over the spaces before and after the pipe -- if there are no spaces it works. I can reproduce the issue with the local server running, no need to generate a static report.

This is surprising to me because there's plenty of places where we pipe templatetags with spaces around the pipe (like [here](https://github.com/ansible-community/ara/blob/7ce19a37c0c42cc0ad2b36b3ebed19bed99ff0df/ara/ui/templates/partials/playbook_results.html#L161-L162)) so I'm mildly confused at it -- it may have to do with being inside an if block ? I'll have learned something today.

Anyway, I suggest switching the condition around so we don't need to test for the length (an empty list is false):

```suggestion
                    {% if result.task.tags %}
                      <span class="badge badge-pill badge-secondary" class="btn btn-secondary" data-toggle="tooltip"  data-placement="right" data-html="true" title="{% include 'partials/tags_list.html' with tags=result.task.tags %}">
                    {% else %}
                      <span class="badge badge-pill badge-secondary invisible" class="btn btn-secondary">
                    {% endif %}
                      {{ result.task.tags | length }} tag{{ result.task.tags | pluralize }}
                      </span>
```

I've also added a ``| pluralize`` filter that automatically adds an "s" if there is >1 tag.

ara/ui/templates/partials/playbook_results.html
```diff
@@ -127,6 +127,13 @@ <h4>
                     {% else %}
                       {% include "partials/result_status_icon.html" with status=result.status %}
                     {% endif %}
+                    {% if result.task.tags | length == 0 %}
+                      <span class="badge badge-pill badge-secondary invisible" class="btn btn-secondary">
+                    {% else %}
+                      <span class="badge badge-pill badge-secondary" class="btn btn-secondary" data-toggle="tooltip"  data-placement="right" data-html="true" title="{% include 'partials/tags_list.html' with tags=result.task.tags %}">
+                    {% endif %}
+                      {{ result.task.tags | length }} tags
+                      </span>
```
dmsimard, line 136:
```suggestion
                    {% if result.task.tags %}
                      <span class="badge badge-pill badge-secondary" class="btn btn-secondary" data-toggle="tooltip"  data-placement="right" data-html="true" title="{% include 'partials/tags_list.html' with tags=result.task.tags %}">
                    {% else %}
                      <span class="badge badge-pill badge-secondary invisible" class="btn btn-secondary">
                    {% endif %}
                      {{ result.task.tags | length }} tags
                      </span>
```

ara/ui/templates/partials/playbook_results.html
```diff
@@ -174,3 +181,14 @@ <h3>No results found matching your query: try <a href="{% url 'ui:playbook' play
...
+  $(function () {
+    $('[data-toggle="tooltip"]').tooltip()
+  })
+</script>
+<style>
+  /* Prevent tooltip flickering when using hover */
+  .tooltip {
+    pointer-events: none;
+  }
+</style>
```
dmsimard, line 194: Could you add the ``<script>`` to host_results.html too ? The tooltip doesn't work there because it's missing.

For the style block, could you move it to [ara.css](https://github.com/ansible-community/ara/blob/master/ara/ui/static/css/ara.css) instead ?

Thanks !

JayV3e:
I move the <style> to ara.css

I didn't know about pluralize, this is pretty sweet. Thanks for the tip!

All suggestions have been applied.

dmsimard:
I should have merged this sooner, I'll tweak it a little bit based on feedback but it'll go in 1.5.8. Thanks !

dmsimard:
Tweaks in this PR: https://github.com/ansible-community/ara/pull/375

PR #285, opened 2021-07-10
zuul: Use fedora 34 nodesets

dmsimard:


dmsimard:
Nevermind, fedora-latest was updated to 34 recently.

PR #290, opened 2021-07-20
docs: remove unused lexer and unpin sphinx

dmsimard:
Docs were failing to build on sphinx>=4 due to the pygments lexer that
we no longer use for the repository since roles have been moved out to
a collection.

Remove it and, if need be, consider using the following in the future:
https://github.com/ansible-community/ansible-pygments


dmsimard:
This will also unblock a fedora package failure to build from source in rawhide that has moved on to sphinx 4: https://bugzilla.redhat.com/show_bug.cgi?id=1977304

PR #293, opened 2021-07-22
api/callback/ui: Add support for tracking task delegation

dmsimard:
ara did not previously record whether or not a task had been
delegated and if so, to which host. Now it does.

This adds a field for it in the result model so we can display it when
browsing a result's details.

From an API perspective, it also adds support for searching results that
may have been delegated to a particular host ID like so:
    /api/v1/results?delegated_to=<id>


noonedeadpunk:
Code looks pretty fair to me

ara/plugins/callback/ara_default.py
```diff
@@ -521,6 +521,11 @@ def _load_result(self, result, status, **kwargs):
         # Retrieve the host so we can associate the result to the host id
         host = self._get_or_create_host(hostname)
 
+        # If the task was delegated to another host, retrieve that too
+        delegated_to = None
+        if result._task.delegate_to and result._task.delegate_to != result._host.get_name():
+            delegated_to = self._get_or_create_host(result._task.delegate_to)
```
dmsimard, line 527: It turns out that ``result._task.delegate_to`` is not "resolved" if the task is a loop and can return a jinja expression instead:
Screenshot from 2021-07-27 14-52-55

Talked to #ansible-devel about it and it's a bug they can fix but we need to find a workaround in the meantime.

dmsimard: Upstream issue: https://github.com/ansible/ansible/issues/75339

dmsimard: Found a workaround and sent a PR to fix it: https://github.com/ansible-community/ara/pull/306

PR #296, opened 2021-07-26
ui: Change page title dynamically to provide context

dmsimard:
The page title was always "ARA Records Ansible" which made it harder to
distinguish particular pages across multiple tabs.

The new titles provide a bit of context and provide summary information.
Uncapitalizing ARA because it's another recursive acronym and the
current logo doesn't capitalize the "ara".


dmsimard:
Before:
Screenshot from 2021-07-25 20-14-06

After:
Screenshot from 2021-07-25 21-07-24

PR #298, opened 2021-07-26
api: rename distincthosts to latesthosts

dmsimard:
This has no incidence on functionality and is just a name change intended
to clarify and streamline the meaning and vocabulary around the feature.

Depending on the context, the words "latest" and "distinct" could have
different meanings or be used interchangeably.

Instead of having the notion of a DistinctHost which has a "latest"
that in reality is a "host", do away with the potential confusion and
name it LatestHost with a "host" reference instead of "latest".


dmsimard:
o/ @hille721 is this alright with you ?

I'm not one for bikeshedding or arguing about names too much (naming things is hard) but after working for a while with it for the new host page, I thought this would be a better fit.

hille721:
@dmsimard : in fact that was also my first name for it, but because the ara hosts are not unique I thought it could be confusing. But if you say it is fine, let's go with it!

But I wondering if it wouldn't be cleaner if we are doing a new database migration. Otherwise everyone who already used the master branch for his ara server will have a broken DB.

hille721 (commented):
If you already using the master branch for your ara server, you will get a broken DB with this change

dmsimard:
> @dmsimard : in fact that was also my first name for it, but because the ara hosts are not unique I thought it could be confusing. But if you say it is fine, let's go with it!

:+1: thanks !

> But I wondering if it wouldn't be cleaner if we are doing a new database migration. Otherwise everyone who already used the master branch for his ara server will have a broken DB.

Yes, I am aware -- running from source off of master isn't recommended or supported because these things can happen when things are in flux and move around before settling when the next release comes out.

No impact for those still on the latest release, 1.5.6.

PR #300, opened 2021-07-26
ui: Fix latest host report ordering by date

dmsimard:
The intent was to order hosts by their latest update with the most
recent at the top: -updated.

However, it turns out that we were sorting LatestHost objects instead of
Host objects. This commits fixes it so we sort based on the date of the
host, not the latesthost reference.


dmsimard:
Fix for the bug troubleshooting there: https://github.com/ansible-community/ara/pull/206/files/019db140c40ab524734e3b11dfc43c8140ef3c05#r676980438

PR #303, opened 2021-07-27
api: filter latesthost dates on the nested host instead

dmsimard:
When querying for created or updated dates for the latesthost endpoint,
filter on the dates of the referenced host instead.

### DIFF
--- a/ara/api/filters.py
+++ b/ara/api/filters.py
@@ -159,7 +159,7 @@ class HostFilter(BaseFilter):
     # fmt: on
 
 
-class LatestHostFilter(BaseFilter):
+class LatestHostFilter(django_filters.rest_framework.FilterSet):
     playbook = django_filters.NumberFilter(field_name="host__playbook__id", lookup_expr="exact")
     name = django_filters.CharFilter(field_name="host__name", lookup_expr="icontains")
     changed__gt = django_filters.NumberFilter(field_name="host__changed", lookup_expr="gt")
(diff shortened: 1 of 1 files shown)


hille721:
@dmsimard understand the PR, but maybe there is also the interest to really filter for the creation of the LatestHost object. 
E.g. you want to know which new hosts are managed by Ansible since last week.

Thus I would suggest to add another filter, let's call it maybe "latesthost_created" where we filter on the created field of the LatestHost.

dmsimard:
@hille721 thanks for chiming in and in fact I thought about that too.

I tried to come up with use cases where someone might want to filter on the latesthost dates instead but didn't find one that didn't circle back to the actual host's dates.

In your example:
> E.g. you want to know which new hosts are managed by Ansible since last week.

This would also work by filtering on the nested host dates, yes ? The LatestHost is more or less just a reference to the host itself.

If nothing else, it bothers me a little bit that someone might expect the field to filter against LatestHost but in fact it filters against the Host even though it's a "special" API endpoint compared to the others to begin with.

hille721:
> This would also work by filtering on the nested host dates, yes ? The LatestHost is more or less just a reference to the host itself.

Mmh, no? The `created` field of the `LatestHost` indicates when a certain hosts showed up the first time in ara. The `created` field of it's related `Host` object shows when this host showed up the last time in ara.
Thus if you want to know which hosts (distinct hosts) shows up the first time in the last week, you have to filter for the `created` field of the `LatestHost` itself.

Btw. that's a good example were the naming "LatestHost" could lead to missunderstanding.

hille721:
> If nothing else, it bothers me a little bit that someone might expect the field to filter against LatestHost but in fact it filters against the Host even though it's a "special" API endpoint compared to the others to begin with.

For me this an argument against this PR? Because now you are filterimg against the `Host` object and not as expected against the `LatestHost` object.

dmsimard:
> > This would also work by filtering on the nested host dates, yes ? The LatestHost is more or less just a reference to the host itself.
> 
> Mmh, no? The `created` field of the `LatestHost` indicates when a certain hosts showed up the first time in ara. The `created` field of it's related `Host` object shows when this host showed up the last time in ara.
> Thus if you want to know which hosts (distinct hosts) shows up the first time in the last week, you have to filter for the `created` field of the `LatestHost` itself.

I see what you mean and it makes sense, you're right.

> Btw. that's a good example were the naming "LatestHost" could lead to missunderstanding.

Yes. Naming things is hard :(

dmsimard:
> > If nothing else, it bothers me a little bit that someone might expect the field to filter against LatestHost but in fact it filters against the Host even though it's a "special" API endpoint compared to the others to begin with.
> 
> For me this an argument against this PR? Because now you are filterimg against the `Host` object and not as expected against the `LatestHost` object.

It is and that's why it bothers me.

I will give this some more thought and try to figure out the right approach given that 1.5.7 is delayed for now due to a bug in Ansible with delegated_to.

dmsimard:
@hille721 I've sent a PR to address your feedback, can you let me know if this approach makes more sense: https://github.com/ansible-community/ara/pull/307 ?

PR #304, opened 2021-07-27
ui: Add a message when no playbooks are found

dmsimard:
We have a similar message elsewhere when no results are found, let's add
one here too.


dmsimard:
There's a large diff due to fixing an indentation mistake but it's really just adding this bit:
https://github.com/ansible-community/ara/pull/304/files#diff-45630eba4e4eac692e43675e4c859cced58e75f8286d58d5aba85d45dc2a2b9aR222-R230

PR #306, opened 2021-07-28
api/callback/ui: Re-implement delegate_to task recording

dmsimard:
After testing the initial implementation in production, it had two issues:
1) A task could potentially be delegated to more than a single host if
   delegate_to is a variable provided by a loop (ex: {{ item }})
2) Inside the callback, result._task.delegate_to could return a variable
   instead of a value when the task is a loop.
   See ansible/ansible issue #75339 for details.

We're addressing the first issue by changing the model and API to expect
delegated_to to be a list -- even if it's empty or just one item.

For the second issue, we found that the value of delegate_to is properly
templated for a task loop from within the v2_runner_item_on_* callback
hooks. The callback was adjusted to leverage these item hooks to get the
proper list of delegated host names.


dmsimard:
Note: this necessarily changes the SQL migration for the delegated_to field.
If users have run the previous version of the migration, it will unfortunately not work unless they restore from a backup prior to running the migrations.

PR #307, opened 2021-07-28
api/ui: restore filtering on latesthost & add nested date filtering

dmsimard:
commit 982f1a84422654ff31a5a7c1e92ea122a9054e98 made it so we could no
longer filter by created or updated for LatestHost which made it
inconsistent with all the other endpoints.

This adds host__ fields to filter on the nested host fields instead.

### DIFF
--- a/ara/api/filters.py
+++ b/ara/api/filters.py
@@ -159,7 +159,7 @@ class HostFilter(BaseFilter):
     # fmt: on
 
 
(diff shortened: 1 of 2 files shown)


hille721:
looks good and much more consistent imo.

dmsimard:
Thanks @hille721 !

PR #308, opened 2021-07-29
api: Update LatestHost when deleting a playbook

dmsimard:
We have a custom perform_destroy that runs when deleting a host via the
API (DELETE /api/v1/hosts/<id>) that takes care of updating the
latesthost entry.

However, because hosts get deleted via the database model
(on_delete=models.CASCADE) when deleting a playbook via the API, the
latesthost entry would get deleted when the host was deleted instead of
being updated.

This adds a very similar perform_destroy for playbooks that iterates
over the hosts and if one of the hosts happens to be a LatestHost, it
will update the entry before deleting.

### DIFF
--- a/ara/api/tests/tests_latesthost.py
+++ b/ara/api/tests/tests_latesthost.py
@@ -82,7 +82,7 @@ class LatestHostTestCase(APITestCase):
 
         first_host = self.client.post("/api/v1/hosts", {"name": "localhost", "playbook": first_playbook.id})
         # Slow this down a bit so we have a second host that is more noticeably "latest"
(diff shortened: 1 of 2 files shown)


dmsimard:
@hille721 FYI

dmsimard:
recheck

hille721:
@dmsimard: ooh..., do we covered then all possibilities? Are the CLI commands also strictly use these endpoints?

Another solution would be to overwrite the `delete` method of the `Host` model directly. Then it doesn't matter anymore how a host gets deleted and the LatestHost will be updated in any case.

Issue #311, opened 2021-07-30
Upgading to ara[server] 1.5.7rc4 will lead to errors with a 1.5.6 (or smaller) callback plugin


hille721:
## What is the issue ?
After upgrading my ara server to 1.5.7rc4 I get following errors if I am using a ara callback version 1.5.6 (or smaller):
On Client: 
```
Failed to post on /api/v1/results
```
On Server:
```
Failed to post on /api/v1/results
```
And no results are stored in the DB:
![image](https://user-images.githubusercontent.com/38468371/127665036-853db9aa-849d-4bab-be29-53b9e3e4e4bd.png)

If I update the client to 1.5.7rc4 it works

## What should be happening ?
The callback plugin should still work for 1.5.6 or smaller versions

hille721:
I found the problem: the API endpoint `/api/v1/results` expects `delegated_to`as mandatory field in the payload. But in previous version it is of course not there. 
Thus we have to made the field optional

hille721:
fixed with #312

PR #312, opened 2021-07-30
Ensure compatibility with previous ara callback versions

hille721:
This PR fixes issue #311.

Through the introduction of the "deleagted_to" field in the result object, the API endpoint `//api/v1/results` expects `delegated_to` as mandatory field in the payload. Because previous ara callback version does not have this ara would not be compatible with older callback versions anymore.

The fix is simple, by only adding a `required=False` in the serializer.


dmsimard:
Thanks for the fix @hille721 !

Issue #323, opened 2021-08-06
/api/v1/results much slower than the other API endpoints in browsable API interfaces


hille721:
## What is the issue ?
The `/api/v1/results` endpoint is much slower than the other ones. In a browser I'm running again and again in 502 - Bad Gateway Errors.

Here a comparison with Google networkanalysis:
 `/api/v1/tasks` (total ~400,000 tasks): ~1s
  `/api/v1/results` (total ~2,000,000 results): ~8s (and around every second time a gateway error)

By using curl, there is not really a significant difference, thus just a GUI problem.

I'm using ara in container behind a nginx proxy in Openshift. I already created my own image with container optimized gunicorn settings (see #322), which solved the problem a bit. But not totally.

## What should be happening ?
Loading   `/api/v1/results` should not take that long.

hille721:
I already found out, that the performance issue is not because of the request itself, it is really just the rendering. And here also not the rendering of the get result, but the rendering of the POST and PUT forms.

Thus with setting `WRITE_LOGIN_REQUIRED: true` you do not have the problem, but you can't do that by running with a proxy.

hille721:
Cool, I have already found a fix for this issue, in which just the html form in the POST/PUT section is disabled, but you can still post data in raw format via the browseable API: https://stackoverflow.com/questions/18587726/django-rest-framework-slow-browsable-ui-because-of-large-related-table

Following has to be done: https://github.com/hille721/ara/commit/2318605dd56c3b1f9e98714354ac62e8a20baace

Then the form will look like this:
![image](https://user-images.githubusercontent.com/38468371/128544804-8c455281-fd36-436a-a9ee-b3554d242f8a.png)
So no html form anymore.

This fix reduces the loading time in my Dev instace from 18.5 s to 3.5 s!

Will submit a PR

dmsimard:
PR is merged, thanks !

PR #324, opened 2021-08-06
Disable html forms in browsable api

hille721:
If you are running ara in production you will collect a lot of data over time. This will slow down the browsable API because of the population of related keys into the dropdowns of the html POST/PUT form. 

This is noticed most strongly in the results API view (`/api/v1/results` or `/api/v1/results/X`) because there we have a lot of related fields (delegated_to, host, task, play, playbook). (compare #323)

Because the browsable API is really nice, we need a solution, where we can still browse the API, and also post data if we want, but withouth this huge performance overload.

This code change is based on a anwser on stackoverflow: https://stackoverflow.com/questions/18587726/django-rest-framework-slow-browsable-ui-because-of-large-related-table and just disables the html forms.
That's how the POST/PUT section is looking then:
image

Performance Test:
DB with following data:
* hosts: 290,569
* tasks: 408,343
* plays: 10,271
* playbooks:  9,795

loading time of `/api/v1/results` :  
* before: ~ 18.5 s
* after: ~ 3.5 s

### DIFF
--- a/ara/api/renderers.py
+++ b/ara/api/renderers.py
@@ -1,0 +1,8 @@
+from rest_framework.renderers import BrowsableAPIRenderer
+
+
+class BrowsableAPIRendererWithoutForms(BrowsableAPIRenderer):
+    """Renders the browsable api, but excludes the forms."""
+
+    def get_rendered_html_form(self, data, view, method, request):
+        return None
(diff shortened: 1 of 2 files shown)


dmsimard:
Works for me! I've reproduced the performance improvement on demo.recordsansible.org and we can also tell the difference just by flipping WRITE_LOGIN_REQUIRED from false to true.

Before we merge, could I ask you to add a comment in the class as to why we are disabling the forms ? Comment doesn't need to be exhaustive but should link to https://github.com/ansible-community/ara/issues/323 where the explanation is.

Thanks for finding the problem *AND* contributing a fix :heart:

PR #327, opened 2021-08-20
cli: add --latest to 'ara host list'

dmsimard:
This mirrors the functionality provided by the hosts page of the
reporting interface.

By specifying --latest, only the latest playbook report will be
returned for each host.


dmsimard:
recheck

Issue #237, opened 2021-03-26
"hardcoding" the controller field as the fqdn (instead of hostname) is not always appropriate


dmsimard:
## What is the issue ?

The callback currently uses ``socket.getfqdn()`` to retrieve the fqdn of the controller:
https://github.com/ansible-community/ara/blob/1371c9860adba8abcfa62b0d96154d672120b60c/ara/plugins/callback/ara_default.py#L296

However, it turns out that the fqdn is a bit much in some cases reported by users such as ``1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa`` on Mac or ``ip-10-0-80-179.us-east-2.compute.internal`` for an AWS VM.

## What should be happening ?

We might want to consider using ``socket.gethostname()`` instead but it would result in controllers being saved under a different name moving forward. I suppose it would be possible to extract the hostname from the fqdn and then update it in a SQL migration but I worry that the "before" and "after" hostnames aren't identical. Need to look into it.

Nudin:
I think the better – more flexible – way of moving on would be to allow to overwrite the controller value via an ara option. That way, one could for example also use this option to set the controller field to the name of the person that did the deployment.

RicardoM17:
Any updates on this? Someone on my team is having this same issue and it is indeed annoying as it clutters ARA quite a bit.

Overwriting based on an env variable would be good enough for our side. Seems like a good feature.

PR #336, opened 2021-11-08
Add a localhost_to_hostname toggle in the ansible callback

dmsimard:
In use cases where playbooks are run against localhost, it might be
useful to record results under the fqdn of the host instead in order to
differentiate task results belonging to different hosts.

TODO:
- integration tests
- docs

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -161,6 +161,33 @@ options:
     ini:
       - section: ara
         key: ignored_files
+  localhost_as_hostname:
+    description:
+        - Associates results to the hostname (or fqdn) instead of localhost when the inventory name is localhost
+        - Defaults to false for backwards compatibility, set to true to enable
+        - This can be useful when targetting localhost, using ansible-pull or ansible-playbook -i 'localhost,'
+        - This helps differentiating results between hosts, otherwise everything would be recorded under localhost.
+    type: boolean
+    default: false
+    env:
+      - name: ARA_LOCALHOST_AS_HOSTNAME
+    ini:
+      - section: ara
+        key: localhost_as_hostname
+  localhost_as_hostname_format:
+    description:
+      - The format to use when recording the hostname for localhost
+      - This is used when recording the controller hostname or when ARA_LOCALHOST_TO_HOSTNAME is true
+      - There are different formats to choose from based on the full (or short) configured hostname and fqdn
+      - Defaults to 'fqdn' (i.e, server.example.org) but can be set to 'fqdn_short' (server)
+      - Other options include 'hostname' and 'hostname_short' which may be suitable depending on server configuration
+    default: fqdn
+    env:
+      - name: ARA_LOCALHOST_AS_HOSTNAME_FORMAT
+    ini:
+      - section: ara
+        key: localhost_as_hostname_format
+    choices: ['fqdn', 'fqdn_short', 'hostname', 'hostname_short']
   callback_threads:
     description:
       - The number of threads to use in API client thread pools
@@ -188,6 +215,7 @@ class CallbackModule(CallbackBase):
     def __init__(self):
         super(CallbackModule, self).__init__()
         self.log = logging.getLogger("ara.plugins.callback.default")
+        self.localhost_hostname = None
         # These are configured in self.set_options
         self.client = None
         self.callback_threads = None
@@ -195,6 +223,8 @@ class CallbackModule(CallbackBase):
         self.ignored_facts = []
         self.ignored_arguments = []
         self.ignored_files = []
+        self.localhost_as_hostname = None
+        self.localhost_as_hostname_format = None
 
         self.result = None
         self.result_started = {}
@@ -216,6 +246,8 @@ class CallbackModule(CallbackBase):
         self.ignored_facts = self.get_option("ignored_facts")
         self.ignored_arguments = self.get_option("ignored_arguments")
         self.ignored_files = self.get_option("ignored_files")
+        self.localhost_as_hostname = self.get_option("localhost_as_hostname")
+        self.localhost_as_hostname_format = self.get_option("localhost_as_hostname_format")
 
         client = self.get_option("api_client")
         endpoint = self.get_option("api_server")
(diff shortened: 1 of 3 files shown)


dmsimard:
What it looks like when running against localhost by default:
Selection_209

And then what it looks like when ``ARA_LOCALHOST_AS_FQDN`` is ``true``:
*(disregard the fact that "localhost.localdomain" is my laptop's FQDN, I'm funny like that)*
Selection_212

hille721:
I like this idea. But might it be also possible to record the hostname instead of the DNS? Imo this would be more consistent, because "normal hosts" are normally also recorded with hostnames only.

dmsimard:
@hille721 so in other words: ``hostname`` instead of ``hostname -f`` ?

We currently use the FQDN for the controller and it has, in fact, been reported as "too verbose" before: https://github.com/ansible-community/ara/issues/237

hille721:
@dmsimard : exactly. E.g., on the Demo page (https://demo.recordsansible.org/hosts) all hostnames are short. Having now FQDNs for localhost tasks would be very inconsistent.

hille721:
also a Ansible controller would be probabbly appears twice in the un

dmsimard:
> @dmsimard : exactly. E.g., on the Demo page (https://demo.recordsansible.org/hosts) all hostnames are short. Having now FQDNs for localhost tasks would be very inconsistent.

Just to clarify, they happen to be short because that's what happened to be their ansible [inventory_hostname](https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html) though, there's no DNS query or resolution involved for those.

If I have an inventory file like this:
```
server.example.org ansible_host=127.0.0.1 ansible_connection=local
```

It'll be saved as ``server.example.org``:
Screenshot from 2021-11-08 17-04-39

Maybe that server's hostname (as configured in the OS) has nothing to do with ``server.example.org``, maybe it's ``bastion.example.org``. Thus: an important distinction between what the host is named after in ansible and the hostname we get when querying the OS \o/

> also a Ansible controller would be probabbly appears twice in the latest host table. One time with FQDN for localhost tasks, one time with short hostname if he is a normal Ansible host as well.

Yes, IMO we should return the same thing for the controller as well as the "translation" for localhost.

So if we say just the short hostname is better (or sufficient?), we should think about changing it for the controller as well (or make it configurable?) like described in https://github.com/ansible-community/ara/issues/237

hille721:
> Just to clarify, they happen to be short because that's what happened to be their ansible inventory_hostname though, there's no DNS query or resolution involved for those.

yes I know that, but I guess th

flowerysong:
> yes I know that, but I guess the most people will using short names for their ansible inventory_hostname.

I would guess that most people will use FQDNs. (Guessing is a lot of fun.)

Regardless of which one is more popular, both are common and need to be supported.

dmsimard:
> > Yes, IMO we should return the same thing for the controller as well as the "translation" for localhost
> 
> right, but haven't meant that. I mean if your Ansible Controller is also in the inventory (with shortname) and you are running a playbook on that, it will come with its short name in the host table. Running now something with `delegate to: localhost` would bring it with FQDN to the host table. Thus its not unique anymore.

Your comment made me think about delegate_to which has some special treatment, i.e:
https://github.com/ansible-community/ara/blob/608006cec32d37878b197c0913c8c20f0f5c77b6/ara/plugins/callback/ara_default.py#L548-L560

Should we also translate ``delegate_to: localhost`` to the hostname/fqdn ?

dmsimard:
> I would guess that most people will use FQDNs. (Guessing is a lot of fun.)
> 
> Regardless of which one is more popular, both are common and need to be supported.

@flowerysong yes, in fact if I recall correctly I had created #237 based on your feedback.

How about we rename the proposed ``ARA_LOCALHOST_TO_FQDN`` to ``ARA_LOCALHOST_TO_HOSTNAME`` and then have something like:
```
options:
  # ...
  localhost_hostname_format:
    description:
      - The format to use when recording the hostname for localhost
      - This is used when recording the controller hostname or when ARA_LOCALHOST_TO_HOSTNAME is true
      - Defaults to 'fqdn' (server.example.org) but can be set to 'short' (server)
    default: fqdn
    env:
      - name: ARA_LOCALHOST_HOSTNAME_FORMAT
    ini:
      - section: ara
        key: localhost_hostname_format
    choices: ['short', 'fqdn']
```

hille721:
> How about we rename the proposed ARA_LOCALHOST_TO_FQDN to ARA_LOCALHOST_TO_HOSTNAME and then have something like

great solution, would completely fits my use cases!

flowerysong:
@dmsimard There are at least four variations people might want to use, some or all of which might be the same depending on configuration.

```
#!/usr/bin/env python3

import socket


print('Reverse lookup: {}'.format(socket.getfqdn()))
print('Reverse lookup short: {}'.format(socket.getfqdn().split('.')[0]))
print('Node name: {}'.format(socket.gethostname()))
print('Node name short: {}'.format(socket.gethostname().split('.')[0]))
```

Results from three different machines:
```
Reverse lookup: ip-10-0-80-100.us-east-2.compute.internal
Reverse lookup short: ip-10-0-80-100
Node name: pandora.x.mail.umich.edu
Node name short: pandora
```

```
Reverse lookup: vereveel.marwnad.com
Reverse lookup short: vereveel
Node name: vereveel.marwnad.com
Node name short: vereveel
```

```
Reverse lookup: merrill
Reverse lookup short: merrill
Node name: merrill
Node name short: merrill
```

dmsimard:
@hille721, @flowerysong I've pushed a new commit which allows for four different hostname formats:
https://github.com/ansible-community/ara/pull/336/commits/92f0cc7999726910718840a5080c9bf08fa419fa#diff-9172751311469aad9e8a49a4d5f962a99f56e2683ecc08332d073c46cdc728a6R177-R190

The hostname (or fqdn) is retrieved in this method:
https://github.com/ansible-community/ara/pull/336/commits/92f0cc7999726910718840a5080c9bf08fa419fa#diff-9172751311469aad9e8a49a4d5f962a99f56e2683ecc08332d073c46cdc728a6R665-R682

Thanks for your feedback, I like this version better :)

dmsimard:
I've pushed an updated commit that cleans up the implementation a little bit. Should be ready to go unless someone has feedback about it.

hille721:
> I've pushed an updated commit that cleans up the implementation a little bit. Should be ready to go unless someone has feedback about it.

I like this version as well. Sorry for the late feedback, I wanted to test it first, but too busy the last days.

dmsimard:
I've fixed the truthy issue and jobs are happy.

@flowerysong this should allow you to adjust the controller hostname so it's shorter. Can you check it out and let me know if it works for you ?

dmsimard:
recheck

dmsimard:
recheck

dmsimard:
recheck

dmsimard:
Looks good to me as is, let's ship it :rocket:

Issue #286, opened 2021-07-13
Container image CI job is broken


dmsimard:
## What is the issue ?

The [CI job](https://dashboard.zuul.ansible.com/t/ansible/build/c5792ed5cc1d407aa392711fad31a4d1) that builds and tests container images is broken and returns the following errors:
```
+ buildah run '' -- /bin/bash -c 'dnf update -y && dnf install -y which python3-pip python3-psycopg2 python3-mysql python3-gunicorn && dnf clean all'
time="2021-07-13T21:19:15Z" level=warning msg="error running newgidmap: exit status 1: newgidmap: write to gid_map failed: Operation not permitted\n"
time="2021-07-13T21:19:15Z" level=warning msg="falling back to single mapping"
time="2021-07-13T21:19:15Z" level=warning msg="error running newuidmap: exit status 1: newuidmap: write to uid_map failed: Operation not permitted\n"
time="2021-07-13T21:19:15Z" level=warning msg="falling back to single mapping"
error reading build container "": error finding build container
time="2021-07-13T21:19:15Z" level=error msg="exit status 125"
+ buildah run --volume /home/zuul/src/github.com/ansible-community/ara:/usr/local/src/ara:z '' -- /bin/bash -c 'pip3 install /usr/local/src/ara/dist/ara-1.5.7.dev22.tar.gz[server]'
time="2021-07-13T21:19:15Z" level=warning msg="error running newgidmap: exit status 1: newgidmap: write to gid_map failed: Operation not permitted\n"
time="2021-07-13T21:19:15Z" level=warning msg="falling back to single mapping"
time="2021-07-13T21:19:15Z" level=warning msg="error running newuidmap: exit status 1: newuidmap: write to uid_map failed: Operation not permitted\n"
time="2021-07-13T21:19:15Z" level=warning msg="falling back to single mapping"
error reading build container "": error finding build container
time="2021-07-13T21:19:15Z" level=error msg="exit status 125"
+ buildah config --env ARA_BASE_DIR=/opt/ara ''
time="2021-07-13T21:19:16Z" level=warning msg="error running newgidmap: exit status 1: newgidmap: write to gid_map failed: Operation not permitted\n"
time="2021-07-13T21:19:16Z" level=warning msg="falling back to single mapping"
time="2021-07-13T21:19:16Z" level=warning msg="error running newuidmap: exit status 1: newuidmap: write to uid_map failed: Operation not permitted\n"
[...]

PR #347, opened 2022-01-18
zuul: Add a post job to build and publish container images

dmsimard:
We used to do this before migrating to this instance of Zuul, see
3053583ddb6e088fb59a9fa8508438a46561d257.

This adds the jobs back in such that container images will be built,
tested and published to dockerhub and quay.io every time a new PR is
merged.

### DIFF
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -47,6 +47,38 @@
     run: tests/with_container_images.yaml
 
 - job:
+    name: ara-container-images-dockerhub
+    parent: ara-integration-base
+    nodeset: fedora-latest-1vcpu
+    description: |
+      Builds ARA API container images with buildah and briefly tests them with podman.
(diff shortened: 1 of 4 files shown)


mike-586:
这是来自QQ邮箱的假期自动回复邮件。您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

dmsimard:
The post job didn't work:
Screenshot from 2022-01-18 10-03-12

It's because we elevated privileges a while back to workaround an issue: https://github.com/ansible-community/ara/commit/f1de47778311c81fc0c8f7f8c72edc591c2eebb6

So the images are built as root and then we try to tag them as an unprivileged user that can't see them.
We'll elevate the privileges for publishing as well for the time being.

We can try dropping privilege escalation once we migrate to another zuul instance (software factory).

dmsimard:
There's an issue that tracked the container image permission problem: https://github.com/ansible-community/ara/issues/286

PR #349, opened 2022-01-18
zuul: escalate privileges when publishing container images

dmsimard:
In f1de47778311c81fc0c8f7f8c72edc591c2eebb6 we added a workaround for an
issue by escalating privileges when building and testing images.

We need to escalate when publishing as well otherwise the unprivileged
user won't see the images that were built prior.


dmsimard:
Related issue: https://github.com/ansible-community/ara/issues/286

PR #353, opened 2022-01-31
zuul: Update config to work with software factory

dmsimard:
We're migrating CI to another instance of Zuul which has slightly
different nodesets and we need to re-encrypt the secrets.


dmsimard:
re-opening to retrigger zuul

Issue #339, opened 2021-11-20
ui: upgrade from bootstrap 4 to bootstrap 5


dmsimard:
## What is the idea ?

One of the vendored static asset for the reporting UI is [bootstrap](https://getbootstrap.com/) via [bootswatch](https://bootswatch.com/) themes: https://github.com/ansible-community/ara/tree/master/ara/ui/static

Unfortunately, bootstrap 5 was still in beta (or alpha?) when the last UI refresh was done and so I picked bootstrap 4 to do it at the time.

bootstrap 5 as well as the bootswatch themes are available now, providing an opportunity to update them.

dmsimard:
I've created a PR to see what it looks like: https://github.com/ansible-community/ara/pull/340

PR #360, opened 2022-02-28
requirements: update django to 3.2 LTS

dmsimard:
We've been recommending django 2.2 LTS and now that 3.2 has been out for
almost a year, we should bump the requirement.

### DIFF
--- a/ara/server/__main__.py
+++ b/ara/server/__main__.py
@@ -18,7 +18,6 @@
 
 import os
 import sys
-from distutils.version import LooseVersion
 
 from django.conf import settings
 
@@ -27,7 +26,6 @@ from ara.setup.exceptions import (
     MissingMysqlclientException,
     MissingPsycopgException,
     MissingSettingsException,
-    Psycopg2VersionException,
 )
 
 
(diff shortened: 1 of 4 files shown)


dmsimard:
I am cautiously optimistic but unless mistaken, it seems like it just works ?
I didn't test exhaustively but the server and interface seem to run fine as well as recording a playbook locally.

dmsimard:
Should be OK now I think but it would need to be rebased on top of https://github.com/ansible-community/ara/pull/361 once it merges due to an unrelated failing linter test.

This is probably a significant enough change to warrant leaving it out of 1.5.8 (which is already long overdue) but it should otherwise be good to land.

dmsimard:
recheck

dmsimard:
> Should be OK now I think but it would need to be rebased on top of #361 once it merges due to an unrelated failing linter test.
> 
> This is probably a significant enough change to warrant leaving it out of 1.5.8 (which is already long overdue) but it should otherwise be good to land.

CI fixes have been done.

Let's keep this for another version bump, it would be a good fit to include with the update from bootstrap 4 to bootstrap 5: https://github.com/ansible-community/ara/issues/339

dmsimard:
Just noticed, we shouldn't forget about updating the documentation links to point to django 3.2 instead of 2.2: https://raw.githubusercontent.com/ansible-community/ara/master/doc/source/api-configuration.rst

Issue #228, opened 2021-03-23
Ara is logging to stdout together with Ansible resulting in weird output


eRadical:
## What is the issue ?
Ara is logging to stdout together w/ Ansible resulting in garbled output hard to read.
If I'm unsetting ara's callbacks Ansible's output is back.

```
$ ansible-playbook ara-breaks-stdout-logging.yaml 
Operations to perform:
  Apply all migrations: admin, api, auth, contenttypes, db, sessions
Running migrations:
  No migrations to apply.

PLAY [Example] *************************************************************************************************************************
2021-03-23 13:54:09,721 INFO ansible: PLAY [Example] *************************************************************************************************************************

TASK [a shell command] *****************************************************************************************************************
2021-03-23 13:54:09,815 INFO ansible: TASK [a shell command] *****************************************************************************************************************
Tuesday 23 March 2021  13:54:09 +0200 (0:00:00.791)       0:00:00.791 ********* 
2021-03-23 13:54:09,831 INFO ansible: Tuesday 23 March 2021  13:54:09 +0200 (0:00:00.791)       0:00:00.791 ********* 
changed: [localhost]
2021-03-23 13:54:10,108 INFO ansible: changed: [localhost]

PLAY RECAP *****************************************************************************************************************************
2021-03-23 13:54:10,151 INFO ansible: PLAY RECAP *****************************************************************************************************************************
localhost                  : ok=1    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
2021-03-23 13:54:10,151 INFO ansible: localhost                  : ok=1    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

Playbook run took 0 days, 0 hours, 0 minutes, 0 seconds
2021-03-23 13:54:10,298 INFO ansible: Playbook run took 0 days, 0 hours, 0 minutes, 0 seconds
Tuesday 23 March 2021  13:54:10 +0200 (0:00:00.466)       0:00:01.258 ********* 
2021-03-23 13:54:10,298 INFO ansible: Tuesday 23 March 2021  13:54:10 +0200 (0:00:00.466)       0:00:01.258 ********* 
=============================================================================== 
2021-03-23 13:54:10,298 INFO ansible: =============================================================================== 
a shell command ----------------------------------------------------------------------------------------------------------------- 0.47s
2021-03-23 13:54:10,298 INFO ansible: a shell command ----------------------------------------------------------------------------------------------------------------- 0.47s
```

[...]

PR #367, opened 2022-03-18
logging: fix no logs when ARA callback set

DataHearth:
Problem:
- Ansible log file contains only warning when ARA's callback plugin is set. 
- User LOGGING configuration was not used 

Context:
As mentioned in issue #228, `Django` logging variable was not defined correctly. It was missing a function call to gather user's logging configuration from `dynaconf` like so:

```python
LOGGING = settings.get("LOGGING", ...)
```

Also, due to default LOGGING configuration, when ARA's ansible callback plugin is set, and `log_path` is set in `ansible.cfg`, this result in fewer logs inside ansible log file. Mostly warnings. It happening because ansible logging system uses the `root` configuration to generate its `FileHandler` logger. cf: https://github.com/ansible/ansible/blob/6f445ca6e5c9c8b85ccc5062e00508c69ca26fde/lib/ansible/utils/display.py#L166-L178

As such, by redefining the root logger in `settings.yaml` by default when no user configuration is passed, results to dumping ansible logs. The explanation outside of the ansible's logger, is due to `Django` configuration injector  https://docs.djangoproject.com/en/4.0/topics/settings/ . It will take the new LOGGING field as default logging configuration and override existing root logger.

Solution:
- Get user's LOGGING configuration by retrieving it from `dynaconf`
- Update default LOGGING configuration by removing root key
- Check if user defined a "root" key inside its configuration and, if so, removing it (and print a warning)

### DIFF
--- a/ara/server/settings.py
+++ b/ara/server/settings.py
@@ -66,7 +66,7 @@ LOGGING = settings.get(
         },
         "loggers": {"ara": {"handlers": ["console"], "level": LOG_LEVEL, "propagate": 0}},
     },
-)
+).to_dict()
 
 if "root" in LOGGING:
     print('[ara] "root" key not allowed in "LOGGING" configuration. Removing...')
--- a/ara/server/settings.py
+++ b/ara/server/settings.py
@@ -48,6 +48,8 @@ BASE_DIR = settings.get("BASE_DIR", BASE_DIR)
 DEBUG = settings.get("DEBUG", False, "@bool")
 
 LOG_LEVEL = settings.get("LOG_LEVEL", "INFO")
+# default logger updated cf issue https://github.com/ansible-community/ara/issues/228
+# and PR https://github.com/ansible-community/ara/pull/367
 LOGGING = settings.get(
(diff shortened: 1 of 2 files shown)


ara/server/settings.py
```diff
@@ -48,37 +48,37 @@
...
+                "stream": "ext://sys.stdout",
+            }
+        },
+        "loggers": {
+            "ara": {"handlers": ["console"], "level": LOG_LEVEL, "propagate": 0}
+        },
     },
-}
-# fmt: on
+)
```
dmsimard, line 72: Could you add a comment with a link to the github issue ?

dmsimard:
I'm strangely not able to reproduce the issue from the integration test jobs, will continue looking but in the meantime the black linter is not happy -- you can reproduce locally with ``tox -e linters``.

You can then re-format the file with ``.tox/linters/bin/black --config .black.toml ara/``.

dmsimard:
Looks good to me, thank you so much for hunting this down and fixing it @DataHearth :tada:

Issue #370, opened 2022-03-19
Can't use ARA with python3.5


DataHearth:
## What is the issue ?

When running ARA with `python3.5`, after playing any playbook, you'll get an error `invalid syntax (base.py, line 143)`.

I've tested with `python3.8` and it's working as intended

**Steps to reproduce:**
```bash
docker run -it --rm python:3.5 bash
python3 -m venv /tmp/ara
source /tmp/ara/bin/activate
pip install --upgrade pip
pip install ansible ara[server]
export ANSIBLE_CALLBACK_PLUGINS="$(python3 -m ara.setup.callback_plugins)"
ansible-playbook some-play.yml
ara playbook list
```

playbook:
```yaml
- name: Demo playbook
  hosts: localhost
  tasks:
    - name: Hello world
      debug:
        msg: Hello, World!
```

Ansible verbose logs:
```bash
[DEPRECATION WARNING]: Ansible will require Python 3.8 or newer on the controller
starting with Ansible 2.12. Current version: 3.5.10 (default, Sep 10 2020, 18:30:47)
[GCC 8.3.0]. This feature will be removed from ansible-core in version 2.12.
Deprecation warnings can be disabled by setting deprecation_warnings=False in
ansible.cfg.
/tmp/ara2/lib/python3.5/site-packages/ansible/parsing/vault/__init__.py:44: CryptographyDeprecationWarning: Python 3.5 support will be dropped in the next release of cryptography. Please upgrade your Python.
  from cryptography.exceptions import InvalidSignature
ansible-playbook [core 2.11.9]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /tmp/ara2/lib/python3.5/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /tmp/ara2/bin/ansible-playbook
  python version = 3.5.10 (default, Sep 10 2020, 18:30:47) [GCC 8.3.0]
  jinja version = 2.11.3
  libyaml = True
No config file found; using defaults
host_list declined parsing /etc/ansible/hosts as it did not pass its verify_file() method
Skipping due to inventory source not existing or not being readable by the current user
script declined parsing /etc/ansible/hosts as it did not pass its verify_file() method
auto declined parsing /etc/ansible/hosts as it did not pass its verify_file() method
Skipping due to inventory source not existing or not being readable by the current user
[...]

PR #372, opened 2022-03-21
Pin dynaconf to <3.0.0 when using python3.5

dmsimard:
dynaconf 3.0.0 dropped support for python3.5 by introducing f-strings
which landed in python3.6 and resulted in ara not working under py35.

Let's pin it in order to do a last release compatible with py35 and it
can be removed later.

Closes: https://github.com/ansible-community/ara/issues/370

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,7 +67,9 @@ server=
     django-filter
     django-health-check
     # dynaconf 3.1.3 had a regression https://github.com/ansible-community/ara/issues/212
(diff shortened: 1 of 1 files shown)


dmsimard:
Tested this and ended up with ``dynaconf==2.2.3`` which (thankfully?) works.

dmsimard:
Appropriate for the upcoming 1.5.8 release, we'll drop support for py35 after.

Issue #346, opened 2022-01-05
A playbook ending with a role import failure marked as successfully completed


dmsimard:
## What is the issue ?

I've bumped into this by running [this playbook](https://github.com/ansible-community/ara-collection/blob/master/tests/with_nginx.yaml) and the output is short enough to include here (but it's also [recorded on the demo](https://demo.recordsansible.org/playbooks/3229.html)):
```
> ansible-playbook -i hosts with_nginx.yaml

PLAY [Deploy and test ARA API with nginx, authentication and SSL enabled] *****************************************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************************************************************************************************************************
ok: [ara-api-server]

TASK [Set ara.example.org to 127.0.0.1] ***************************************************************************************************************************************************************************************************************************************************************************************
changed: [ara-api-server]

TASK [Ensure directory for SSL certificate and private key exists] ************************************************************************************************************************************************************************************************************************************************************
changed: [ara-api-server] => (item=/etc/ssl/ara.example.org)
ok: [ara-api-server] => (item=/etc/ssl/ara.example.org)

TASK [Set path to selfsigned openssl configuration] ***************************************************************************************************************************************************************************************************************************************************************************
ok: [ara-api-server]

TASK [Template an openssl configuration file] *********************************************************************************************************************************************************************************************************************************************************************************
changed: [ara-api-server]

TASK [Generate private key and self-signed certificate] ***********************************************************************************************************************************************************************************************************************************************************************
changed: [ara-api-server]

TASK [Set up the API with the ara_api Ansible role] ***************************************************************************************************************************************************************************************************************************************************************************
ERROR! the role 'ara_api' was not found in /home/dmsimard/dev/git/ansible-community/ara-collection/tests/roles:/home/dmsimard/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:/home/dmsimard/dev/git/ansible-community/ara-collection/tests

The error appears to be in '/home/dmsimard/dev/git/ansible-community/ara-collection/tests/with_nginx.yaml': line 90, column 15, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

      include_role:
        name: ara_api
              ^ here

PLAY RECAP ********************************************************************************************************************************************************************************************************************************************************************************************************************
ara-api-server             : ok=6    changed=4    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

[...]

PR #374, opened 2022-03-23
callback: add some missing callback hook debugging

dmsimard:
When troubleshooting callback or playbook execution issues, it can be
useful and relevant to see which hook is currently being run.


dmsimard:
I realized these were missing when troubleshooting https://github.com/ansible-community/ara/issues/346

PR #375, opened 2022-03-25
ui: Move task tags to their own column

dmsimard:
They don't take up too much width, let's put them there and include just
the number.


dmsimard:
Before:
Screenshot from 2022-03-24 20-46-50

After:
Screenshot from 2022-03-24 20-43-47

As a side effect, the status doesn't look misaligned when there's no tags.

PR #381, opened 2022-04-01
Pin updates

dmsimard:


setup.cfg
```diff
@@ -67,11 +67,11 @@ server=
...
     # dynaconf 3.1.3 had a regression https://github.com/ansible-community/ara/issues/212
-    # dynaconf dropped support for py35 at version 3.0.0: https://github.com/ansible-community/ara/issues/370
-    dynaconf[yaml]>=3.0.0,!=3.1.3,<4.0.0;python_version>='3.6'
-    dynaconf[yaml]<3.0.0;python_version<'3.6'
+    dynaconf[yaml]>=3.0.0,!=3.1.3,<4.0.0
     tzlocal<3.0
-    whitenoise
+    # whitenoise 6.0.0 dropped support for py36
+    whitenoise<6.0.0;python_version=='3.6'
+    whitenoise;python_version>'3.6'
```
DataHearth, line 74: @dmsimard shoudn't this version pinned to python 3.7 ? [c.f](https://github.com/evansd/whitenoise/blob/c82f9a3606074d53063243308c2a9afdbe0195cb/setup.cfg#L40)
```suggestion
    whitenoise;python_version>='3.7'
```

dmsimard: @DataHearth I will double check and test to make sure, though my understanding is that ``>3.6`` and ``>=3.7`` are functionally equivalent in this context.

DataHearth: sure !

dmsimard: I've rebased with your suggestion included, let's go with that. Thanks :)

dmsimard:
I screwed up somewhere in the rebase, oops.

I cleaned it up and also formalized >=3.6.

dmsimard:
FYI @DataHearth and @gaby, I've also added an explicit requirement on py>=3.6 -- it was implicit due to django but now it's explicit.

gaby:
> FYI @DataHearth and @gaby, I've also added an explicit requirement on py>=3.6 -- it was implicit due to django but now it's explicit.

Looks good! Question, since you added py3.9, do you know if ara would also work with py3.10 ?

dmsimard:
> > FYI @DataHearth and @gaby, I've also added an explicit requirement on py>=3.6 -- it was implicit due to django but now it's explicit.
> 
> Looks good! Question, since you added py3.9, do you know if ara would also work with py3.10 ?

3.10 is still relatively new and I haven't tested it myself so I'm not sure.
Once upon a time we had different CI jobs to test every version of python explicitly but it made for a large, complicated (and expensive) matrix between that, the version of ansible we tested and the linux distro.

We could bring a limited amount of coverage back so we have better visibility.

PR #392, opened 2022-07-02
Add support for configuring `HttpClient` TLS client certificate and CA

devon-mar:
This PR adds support for configuring a TLS client certificate and CA.

### DIFF
--- a/ara/clients/utils.py
+++ b/ara/clients/utils.py
@@ -10,6 +10,8 @@ def get_client(
     timeout=30,
     username=None,
     password=None,
+    cert=None,
+    key=None,
     verify=True,
     run_sql_migrations=True,
 ):
@@ -20,6 +22,10 @@ def get_client(
     if username is not None and password is not None:
         auth = HTTPBasicAuth(username, password)
 
+    cert = None
+    if cert is not None and key is not None:
+        cert = (cert, key)
+
     if client == "offline":
         from ara.clients.offline import AraOfflineClient
 
@@ -27,7 +33,7 @@ def get_client(
     elif client == "http":
         from ara.clients.http import AraHttpClient
 
-        return AraHttpClient(endpoint=endpoint, timeout=timeout, auth=auth, verify=verify)
+        return AraHttpClient(endpoint=endpoint, timeout=timeout, auth=auth, cert=cert, verify=verify)
     else:
         raise ValueError("Unsupported API client: %s (use 'http' or 'offline')" % client)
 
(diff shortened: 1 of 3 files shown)


dmsimard:
Hi @devon-mar and thanks for the PR.

Sorry for the delay, I've been busy and sidetracked elsewhere.

We are long overdue for a new release in ara and I'll make sure to review your PR before then so it can be included.

ara/clients/utils.py
```diff
@@ -20,14 +22,18 @@ def get_client(
     if username is not None and password is not None:
         auth = HTTPBasicAuth(username, password)
 
+    cert = None
```
dmsimard, line 25: Is this right ? get_client would default ``cert`` to ``None`` and this line would set it to ``None`` even if there is another value passed.

dmsimard:
Honestly, this is great work. Fantastic.

Have only a question but otherwise I think it's good to go.

I'd like to add an integration test for it.
We have playbook in the ansible collection that spins up an nginx in front of the ara server with a self-signed certificate: https://github.com/ansible-community/ara-collection/blob/master/tests/with_nginx.yaml

It turns out there was a TODO to make the callback trust the certificate :p
https://github.com/ansible-community/ara-collection/blob/c19b0474aff31a9bd751dc5f74f4ac201eb4d478/tests/vars/nginx_tests.yaml#L23

ara/clients/utils.py
```diff
@@ -20,14 +22,18 @@ def get_client(
...
+    if cert is not None and key is not None:
+        cert = (cert, key)
```
dmsimard, line 27: My understanding is that they must both be set for it to work but if only one is defined, it would silently do nothing.

Could we raise an exception if either are set but not both ? It can be similar to the validation we're doing for the API client just below.

ara/clients/utils.py
devon-mar: Nice catch! My intention was just to set a default value to be passed to `AraHttpClient`.

ara/clients/utils.py
devon-mar: Yes, both must be set. 

I think the exception would be better rather than silently ignoring it. I'll change that.

devon-mar:
Sorry for the delay! 

I'll make the suggested changes from the review. I'll also try to add an integration test though I'm haven't used Zuul before.

dmsimard:
recheck

dmsimard:
It works, tested via https://github.com/ansible-community/ara-collection/pull/57 and otherwise looks good to me.

Thank you for your patience and contribution @devon-mar, great collaboration on this one :)

PR #403, opened 2022-08-04
CLI: Add support for CA and client cert

dmsimard:
We recently added support for a custom CA and client cert in the
callback, this adds it to the CLI as well.


dmsimard:
recheck

dmsimard:
I've fixed the CI in https://github.com/ansible-community/ara/pull/404 -- I'll wait until monday to merge it and then this might good to go after.

dmsimard:
recheck

Issue #331, opened 2021-09-27
Thoughts and ideas about host index page improvements


hille721:
Hi @dmsimard,

after some weeks/months with the new host index page view, here my thoughts and ideas for improvements:

## thoughts
* the playbook status icon is to prominent in column one. I'm always confused if I see there a red icon, but the host was running green. Also there is the question if we anyhow need this column a the host view.
* same thing about playbook duration. On the host view I'm not really intertested to see this. Imo we should highlight more host specific thins instead
* playbook name is missing. You have to click on the report of one host, to see which playbook it is
* "Last update" and "Last results" are misleading column tiles if "Include all playbook reports" is ticked

## ideas
* reordering of the columns, imo the hostname and last result columns should be highlighted more prominent, eg, by putting them into the beginning
* adding playbook name/path to the view
* dropping playbook status and playbook duration column
* addinng possibility to filer for playbook name/path and subset
* filtering for host facts
* renaming "Last update" and "Last results" to "Update"/"Results"
* adding ansible groups to the view (there I have no idea, if this is technical possible, I think currently the groups in which a host is, is not reported?), but it would be an awesome feature

hille721:
the idea of host fact searching and filtering is also already tracked in #297

dmsimard:
Hey @hille721, thanks for the feedback <3

I need to focus on AnsibleFest and Contributor Summit things for now, let me get back to you with my thoughts when I have a chance.

hille721:
@dmsimard, finally I had some time to look into this and did some changes how I would love too see the host index page:

![image](https://user-images.githubusercontent.com/38468371/182858248-e7501c5b-738d-4b36-8bfc-ae253a6dae3c.png)

* removed playbook status and playbook duration columns
* added playbook name / path and make it searchable (in fact the search is just for the playbook id, not the playbook name. Could be improved)

Addding playbook name / path is btw really easy with `host.playbook.name` / `host.playbook.path`. I would wish to add at least this column to the next release, because this is really missing (also feedbacked by others in my company).

dmsimard:
@hille721 thanks for working on this. Here's some thoughts.

I think I can concede that the playbook status is not immediately relevant and can even be confusing relative to the status of the tasks that the host ran but I'd keep the playbook duration. It's useful to be able to see how long playbooks with that host took.

I wonder about the order of the columns, especially when comparing the playbook page, the hosts page and the new tasks page.
For easy comparison:

---

# Playbooks
[...]

PR #405, opened 2022-08-05
Host index improvements

hille721:
Improvements on host index site. Background: #331 

* filter possibility for playbook name and path
* playbook name column in table

image

### DIFF
--- a/ara/ui/templates/host_index.html
+++ b/ara/ui/templates/host_index.html
@@ -27,6 +27,18 @@
               </div>
             </div>
             <div class="form-row">
+              <div class="form-group col-xl-8">
+                <label for="playbook_name" {% if request.GET.playbook_name %}style="font-weight:bold;" {% endif %}>Playbook name</label>
+                <input type="text" class="form-control" id="playbook_name" name="playbook_name" placeholder="ex: install server" value="{{ search_form.playbook_name.value | default_if_none:'' }}">
+              </div>
+            </div>
+            <div class="form-row">
+              <div class="form-group col-xl-8">
+                <label for="playbook_path" {% if request.GET.playbook_path %}style="font-weight:bold;" {% endif %}>Playbook path</label>
+                <input type="text" class="form-control" id="playbook_path" name="playbook_path" placeholder="ex: /path/to/playbook.yml (or) playbook.yml" value="{{ search_form.playbook_path.value | default_if_none:'' }}">
+              </div>
+            </div>
+            <div class="form-row">
               <div class="form-group col-xl-5">
                 <div class="form-check form-check-inline">
                   <input class="form-check-input" type="checkbox" id="latest" value="false" name="latest" {{ checkbox_status }}>
@@ -114,7 +126,7 @@
           <tr style="height:50px;">
             <th class="text-center" title="Link to the host report">Report</th>
             <th class="text-center" title="Status of the tasks for this host">Task status</th>
-            <th title="Last update">Last update {% include "partials/sort_by_date.html" with arg="updated" %}</th>
+            <th title="Date">Date {% include "partials/sort_by_date.html" with arg="updated" %}</th>
             <th title="Duration of the last playbook the host was involved in">Duration</th>
             <th title="Name of the inventory host">Host name {% include "partials/sort_by_name.html" %}</th>
             <th title="Name (or path) of the playbook">Playbook name (or path)</th>
--- a/ara/ui/templates/host_index.html
+++ b/ara/ui/templates/host_index.html
@@ -112,12 +112,12 @@
       <table class="table table-sm table-hover" id="hosts">
         <thead>
           <tr style="height:50px;">
-            <th class="text-center" title="Status of the playbook">Playbook</th>
             <th class="text-center" title="Link to the host report">Report</th>
+            <th class="text-center" title="Status of the tasks for this host">Task status</th>
             <th title="Last update">Last update {% include "partials/sort_by_date.html" with arg="updated" %}</th>
-            <th title="Duration of the last playbook the host was involved in">Playbook duration</th>
+            <th title="Duration of the last playbook the host was involved in">Duration</th>
             <th title="Name of the inventory host">Host name {% include "partials/sort_by_name.html" %}</th>
-            <th title="Number of task results and their statuses">Last results</th>
+            <th title="Name (or path) of the playbook">Playbook name (or path)</th>
             <th title="Host facts">Host facts</th>
           </tr>
         </thead>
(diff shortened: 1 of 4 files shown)


dmsimard:
@hille721 looks good to me at a first glance -- I'm still trying to fix the unrelated CI issue with the container images.

ara/ui/templates/host_index.html
```diff
@@ -26,6 +26,18 @@
                 <input type="text" class="form-control" id="name" name="name" placeholder="ex: localhost (or) hostname.fqdn" value="{{ search_form.name.value | default_if_none:'' }}">
```
dmsimard, line 26: Should we disambiguate ``name`` here and use ``host_name`` instead ?

hille721: do you mean the label or the the appearances in the code?

hille721:
> @hille721 looks good to me at a first glance -- I'm still trying to fix the unrelated CI issue with the container images.

@dmsimard : in #331 you had a proposal to drop the task icons with zero. I would include this in this PR as well. If you push this to your branch I can rebase this one

ara/ui/templates/host_index.html
dmsimard: Both, I think ? It becomes useful because we now have two names, the host name and the playbook name (if there is one).

For the label it could be "Host name" and the variable host_name ?

dmsimard:
> > @hille721 looks good to me at a first glance -- I'm still trying to fix the unrelated CI issue with the container images.
> 
> @dmsimard : in #331 you had a proposal to drop the task icons with zero. I would include this in this PR as well. If you push this to your branch I can rebase this one

oof, I don't think I committed that to a branch but I'm sure I could come up with something. Hang on.

dmsimard:
@hille721 https://github.com/dmsimard/ara/commit/17ad2ee3621efbe42b120ae51b2f8c40ef5ba86c works for me, I don't think we need something more than that.

ara/ui/templates/partials/host_index_row.html
```diff
@@ -1,51 +1,52 @@
...
-    </a>
+  {% else %}
+  <td title="Search results for {{ host.playbook.path }}">
+    {% if not static_generation %}
+      <a href="{% url 'ui:host_index' %}?playbook_path={{ host.playbook.path }}">{{ host.playbook.path | truncatepath:50 }}</a>
+    {% else %}
+      {{ host.playbook.path | truncatepath:50 }}
     {% endif %}
+  {% endif %}
   </td>
```
dmsimard, line 50: I was leveraging your work to do the same thing in the new tasks page (thanks again!) and was briefly confused due to being two ``<td>``'s but only a single ``</td>``. I understand why it isn't necessary but for the sake of clarity I'll duplicate the </td> inside the if/else block.

ara/ui/templates/host_index.html
dmsimard: Oh, I remembered about this when implementing the task filter and I think I see what you mean.

I think the filter for the host's name should remain "name" but the label and query argument be host_name.
Does that make sense ?

dmsimard: Yeah nevermind, let's do just the label, it's simpler that way :p

ara/ui/templates/partials/host_index_row.html
```diff
@@ -1,51 +1,61 @@
...
+  <td class="text-nowrap">{{ host.playbook.duration | format_duration }}</td>
+  <td>
+    <a {% if not static_generation %}href="{% url 'ui:host_index' %}?name={{ host.name }}&latest=false" title="Search results for {{ host.name }}"{% endif %}>
+      {{ host.name }}
     </a>
+  </td>
+  {% if host.playbook.name is not None %}
+  <td title="Search results for {{ host.playbook.name }}">
+    {% if not static_generation %}
+      <a href="{% url 'ui:host_index' %}?playbook_name={{ host.playbook.name }}">{{ host.playbook.name }}</a>
```
dmsimard, line 46:
```suggestion
      <a href="{% url 'ui:host_index' %}?playbook_name={{ host.playbook.name }}&latest=false">{{ host.playbook.name }}</a>
```

ara/ui/templates/partials/host_index_row.html
```diff
@@ -1,51 +1,61 @@
...
+    {% else %}
+      {{ host.playbook.name }}
+    {% endif %}
+  </td>
+  {% else %}
+  <td title="Search results for {{ host.playbook.path }}">
+    {% if not static_generation %}
+      <a href="{% url 'ui:host_index' %}?playbook_path={{ host.playbook.path }}">{{ host.playbook.path | truncatepath:50 }}</a>
```
dmsimard, line 54:
```suggestion
      <a href="{% url 'ui:host_index' %}?playbook_path={{ host.playbook.path }}&latest=false">{{ host.playbook.path | truncatepath:50 }}</a>
```

dmsimard:
Tested it and it looks good!

I've added two suggestions to include ``&latest=false`` for the playbook name and path because I think the intent would be to return every host result for that playbook, not just the latest one.

What do you think ?

Other than that it's ready to land IMO.

hille721:
mmh good question. I think it can also make sense to get only latest hosts. E.g. to see the last results of a specific playbook. 

But we could make it dependt if `Include all playbook reports` is set or not. What do you thing about that?

ara/ui/templates/partials/host_index_row.html
hille721: added `&latest={{ request.GET.latest | default:'true' }}`. Thus if _Include all playbook reports_ was set before it will be considered in the link as well. 
Does this is fine for you as well?

ara/ui/templates/partials/host_index_row.html
hille721: added `&latest={{ request.GET.latest | default:'true' }}`. Thus if _Include all playbook reports_ was set before it will be considered in the link as well. 
Does this is fine for you as well?

dmsimard:
> mmh good question. I think it can also make sense to get only latest hosts. E.g. to see the last results of a specific playbook.
> 
> But we could make it dependend if _Include all playbook reports_ is set or not. What do you thing about that?
> 
> Edit: with [459b525](https://github.com/ansible-community/ara/pull/405/commits/459b5250803e15fb152fd7328efd5b9d8c22a46b) I added `&latest={{ request.GET.latest | default:'true' }}` to the link to take into account if _Include all playbook reports_ was set or not. Does this make sense to you?

That makes sense.

dmsimard:
Looks good to me, this is a good improvement on the hosts page. Thanks @hille721 :)

PR #408, opened 2022-08-08
ui: ability to search by host or task name

### LINKED ISSUE #243: ui: ability to search by host or task name instead of host ID and task ID
## What is the idea ?

When searching in a playbook task's results, we have host ID and task ID fields:
Screenshot from 2021-04-01 10-14-30

These are not really meant for humans because they are not expected to know the "internal" IDs given by ara to hosts and tasks -- it would be better if users could search by host and task names instead.

I've explained why things are the way they are in a [previous comment](https://github.com/ansible-community/ara/pull/221#issuecomment-800464650) but in a nutshell, it's because we are searching task results and these are associated to task IDs and host IDs, not their names.

In that comment, I suggested a possible approach:
> I haven't given this much thought yet but maybe under the hood we could search for a hostname for that playbook (``/api/v1/hosts?playbook=<id>&name=foo``), get it's ID and then filter with that -- but strictly API speaking, we cannot search a result by hostname right now.

It's worth noting that the current implementation is not useless, though. It allows the UI to provide links on things like host names and task names where clicking on them will search for their IDs.

hille721:
Ability to search for host or task name instead of id (fixes #243)
image

Furthermore this PR introduces host and task name filters on the playbook index as well
image

While the taskfilter supports pattern search (`icontains`), the host filter expects case insensitve exact expression (`iexact`). Because imo that's how most will use it.

### DIFF
--- a/ara/api/filters.py
+++ b/ara/api/filters.py
@@ -184,8 +184,10 @@ class LatestHostFilter(BaseFilter):
 class ResultFilter(DateFilter):
     playbook = django_filters.NumberFilter(field_name="playbook__id", lookup_expr="exact")
     task = django_filters.NumberFilter(field_name="task__id", lookup_expr="exact")
+    task_name = django_filters.CharFilter(field_name="task__name", lookup_expr="icontains")
     play = django_filters.NumberFilter(field_name="play__id", lookup_expr="exact")
     host = django_filters.NumberFilter(field_name="host__id", lookup_expr="exact")
+    host_name = django_filters.CharFilter(field_name="host__name", lookup_expr="iexact")
     delegated_to = django_filters.NumberFilter(field_name="delegated_to__id", lookup_expr="exact")
     changed = django_filters.BooleanFilter(field_name="changed", lookup_expr="exact")
     status = django_filters.MultipleChoiceFilter(
--- a/ara/ui/forms.py
+++ b/ara/ui/forms.py
@@ -20,8 +20,8 @@ class PlaybookSearchForm(forms.Form):
 
 
 class ResultSearchForm(forms.Form):
-    host = forms.CharField(label="Host id", max_length=10, required=False)
-    task = forms.CharField(label="Task id", max_length=10, required=False)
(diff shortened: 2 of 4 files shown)


hille721:
recheck

dmsimard:
@hille721 I finally got around to testing this and it works great, I'm happy that it turned out to be simple after all \o/

This is almost good to merge, two things:
- We should be able to search by task name under the hosts' task results too, here: https://github.com/ansible-community/ara/blob/e3521963bf56c349422d7ef9c5fd8f2018867489/ara/ui/templates/partials/host_results.html#L58-L61
- Although it works, I don't believe the host and task name search fields belong on the playbook index search box (I had commented about this in the [issue](https://github.com/ansible-community/ara/issues/243#issuecomment-1207298842)). We should be fine with the hosts and tasks page for searching about them.

dmsimard:
If you see in this gif, I click on the report link to filter the results by task (id) in the playbook report page: https://github.com/ansible-community/ara/pull/342#issuecomment-1242797898

I guess we need to keep the ability to filter by ID because we do the same thing from the hosts page ?

hille721:
> Although it works, I don't believe the host and task name search fields belong on the playbook index search box (I had commented about this in the issue). We should be fine with the hosts and tasks page for searching about them.

Yeah you probably right. With the tasks / hosts page this is not really needed. 

> If you see in this gif, I click on the report link to filter the results by task (id) in the playbook report page: #342 (comment)
I guess we need to keep the ability to filter by ID because we do the same thing from the hosts page ?

I would say now, we can filter for the task id in the playbook report via query parameter but showing the human friendly name in the filter box. 
Will check that tomorrow.

hille721:
changes made as requested by @dmsimard :

* search by task name under the hosts' task results too & updated the "task link"
* removed host and task name filters out of playbook section

dmsimard:
It works well on my end and it's more straightforward for humans too.

Thanks for for contribution @hille721 :)

PR #413, opened 2022-08-25
Actually return a 405 response like we do with 404.

apollo13:


dmsimard (approved):
LGTM, thanks @apollo13 !

PR #414, opened 2022-08-25
Improved WSGI app selection and hardcoded db backend when distributed sqlite is in use.

apollo13:
@dmsimard This is a follow up to the previous PR. Let me explain the rationale for it:

* I changed `WSGI_APPLICATION` to be static and then do the switch in `wsgi.py`. This has the benefit that the docker container that runs `gunicorn ara.server.wsgi` works out of the box. Accessing settings in `wsgi.py` is okay because `get_wsgi_application()` does it as well.

* I hardcoded `ara.server.db.backends.distributed_sqlite` as ENGINE if distributed sqlite is enabled because there is no way any other database engine would be valid for it. The way I have written it also prevents `ara.server.db.backends.distributed_sqlite` from being written into settings.yml, so one can turn `DISTRIBUTED_SQLITE` on & off at will without having to change the engine in settings to match…

### DIFF
--- a/ara/server/wsgi.py
+++ b/ara/server/wsgi.py
@@ -7,6 +7,7 @@ import os
 from ara.setup.exceptions import MissingDjangoException
 
 try:
+    from django.conf import settings
     from django.core.handlers.wsgi import get_path_info, get_script_name
     from django.core.wsgi import get_wsgi_application
 except ImportError as e:
@@ -17,7 +18,7 @@ os.environ.setdefault("DJANGO_SETTINGS_MODULE", "ara.server.settings")
 logger = logging.getLogger(__name__)
 
 # The default WSGI application
-application = get_wsgi_application()
+default_application = get_wsgi_application()
 
 
 def handle_405(start_response):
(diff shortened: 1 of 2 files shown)


ara/server/wsgi.py
```diff
@@ -86,6 +86,9 @@ def distributed_sqlite(environ, start_response):
 
     local_storage.db_path = db_file
     try:
-        return application(environ, start_response)
+        return default_application(environ, start_response)
     finally:
         del local_storage.db_path
+
+
+application = distributed_sqlite if settings.DISTRIBUTED_SQLITE else default_application
```
dmsimard, line 94: Linter is complaining about no line at the end of file here

dmsimard:
Looks good to me at first glance but I'd like to test it in practice first. Let me get back to you on that.

apollo13:
Sure, take all time you need -- I can easily work around those issues. Don't hesitate to ask if there are any questions.

dmsimard:
I tested it locally and it works for me -- just for good measure, I tested it with various configurations and backends via https://github.com/ansible-community/ara-collection/pull/58 and it didn't pick up on any issues.

Thanks @apollo13 !

PR #416, opened 2022-09-05
Add support for recording usernames

### LINKED ISSUE #415: Log the username on the controller
Hi,

First, thanks for Ara! We are working a lot with Ansible and this seems to fix not one but several issues we had in different setups.

One I'm still thinking about how to fix it with Ara is that we need to know *who* started a run of `ansible-playbook` in one project. If there was a convenient solution Ara would be perfect for us.

Here's the outline of what I'm talking about: We are working with a another team in a common cloud infrastructure. So there's their users as well as our users working on the same bastion host. Everyone has a local copy of common Ansible code (including playbooks, `ansible.cfg`, inventory etc.) in their home on the bastion host. When we change something, we change it in our local copy and contribute into common git repositories. These repositories are then checked out on a similar bastion host in another cloud environment for production. So we know exactly who changed what when in the Ansible code. But we do not know, who ran which playbook in production.

For documentation reasons we want to know who changed when what in production, in other words, who ran which playbook with what result in production. Ara is perfect for that, the only thing I can't find is the information of the user on the Ansible controller who started the playbook. My workaround would be to put it into `default_labels` but I hope there's a better one.

 Just to make it clear: This not about a full blown auditing solution. Yes, that would also be welcome but I guess that would be beyond the scope of this feature request. A prominent information about the username / Gecos of who started a run on the controller would be totally sufficient.

DanOPT:
Add support for recording executing usernames who ran ansible playbooks. 

Fixes #415 

screenshot_1


mocdaniel:
# Issues tackled with this PR

Sorry for repeatedly triggering the pipeline - I misinterpreted the 'local testing' docs. Right now, everything looks fine so far - users issuing a `Playbook` run get recorded as `executor` internally, this field can be used for filtering via UI and CLI.

We also fixed a duplicated CLI argument when issuing `ara record list` which prevented us from building the docs using `tox -e docs`:

`--key` was declared as argument for both, a certificate key for authentication matters (I suppose) and as record key. I adjusted the latter's name to `--recordkey`.

# Open ToDos (cc @DanOPT )

* Check if we need to update the official docs (e.g. regarding the adjusted CLI argument name)
* Double-check if error-handling needs to be put in place for the changes committed to `ara/plugins/callback/ara_default.py` (see also TODO comment)

# Up for discussion

We also thought about introducing a third Navbar entry to the Web UI called `Users` or `Executors` which would group recorded `Playbook` runs by executing user, but wanted to discuss necessity and approach with the maintainers first. Our questions are:

* Is such a contribution needed/wanted right now?
* If so, would you like us to provide this contribution within the scope of this PR (as it is closely tied to the introduction of logging executing users) or as a separate PR building on top of this one.

ara/cli/record.py
```diff
@@ -31,8 +31,8 @@ def get_parser(self, prog_name):
             help=("List records for the specified playbook"),
         )
         parser.add_argument(
-            "--key",
-            metavar="<key>",
+            "--recordkey",
+            metavar="<recordkey>",
```
dmsimard, line 35: Could you explain this issue ? The name of the argument is key and it's what gets passed on to the query later on around line 90: https://github.com/ansible-community/ara/pull/416/files#diff-34f949807054e46caefa740bb513e676fe971ac4b9f71ac109bf432b5de4679bR90-R91

```
> ara record list --playbook 2931
+-----+--------------+------+----------+-----------------------------+
|  id | key          | type | playbook | updated                     |
+-----+--------------+------+----------+-----------------------------+
| 852 | list_as_text | text |     2931 | 2021-02-25T17:01:13.791119Z |
| 851 | somedict     | dict |     2931 | 2021-02-25T17:01:12.255127Z |
| 850 | somelist     | list |     2931 | 2021-02-25T17:01:10.659884Z |
| 849 | bar          | json |     2931 | 2021-02-25T17:01:09.095688Z |
| 848 | foo          | url  |     2931 | 2021-02-25T17:01:07.608241Z |
| 847 | notype       | text |     2931 | 2021-02-25T17:01:04.939126Z |
+-----+--------------+------+----------+-----------------------------+

> ara record list --playbook 2931 --key notype
+-----+--------+------+----------+-----------------------------+
|  id | key    | type | playbook | updated                     |
+-----+--------+------+----------+-----------------------------+
| 847 | notype | text |     2931 | 2021-02-25T17:01:04.939126Z |
+-----+--------+------+----------+-----------------------------+
```

dmsimard: Oh, I was taking a quick glance at the code and missed your comment about it.

I'm not sure whether this is the right fix but we can investigate :p

dmsimard:
Hi @DanOPT and @mocdaniel and thanks for working on this PR :)

It looks good at first glance, some thoughts:

> Sorry for repeatedly triggering the pipeline

No need to apologize, the CI jobs are meant to be run -- the volume is low, the jobs are quick and it's fine so long as we do not abuse :)

I already touched the key argument topic and here's some other thoughts:

> Double-check if error-handling needs to be put in place for the changes committed to ara/plugins/callback/ara_default.py (see also TODO comment)

I think the question we could ask ourselves if whether it is fatal if the callback plugin is somehow unable to retrieve the hostname and the username. I haven't seen it happen but I suppose there could be a case where, somehow, we can't get the hostname properly and then because we [raise an exception](https://github.com/ansible-community/ara/blob/7f8b8367e0ec5083bab836422025f699780cb9f7/ara/plugins/callback/ara_default.py#L739-L741), that prevents the playbook from being recorded at all.

Looking at it from that angle, I'm not sure whether it should really be fatal since we could probably very well record 99% of the remainder of the playbook. We could make the fields nullable and default to None if we can't recover it instead ?

In the UI or CLI, we could decide whether to show anything at all if we don't have a value, write ``n/a`` or something.

> We also thought about introducing a third Navbar entry to the Web UI called Users or Executors which would group recorded Playbook runs by executing user, but wanted to discuss necessity and approach with the maintainers first. Our questions are:

>    Is such a contribution needed/wanted right now?
>    If so, would you like us to provide this contribution within the scope of this PR (as it is closely tied to the introduction of logging executing users) or as a separate PR building on top of this one.

I think this is an interesting idea but we need to think a little bit about how to implement it so we get it right.

We need a bit of the bigger picture so let me explain, currently there is:
- the main index (though it could live at /playbooks) which is somewhat analogous to /api/v1/playbooks
- the hosts page which is somewhat analogous to /api/v1/hosts
- [a work in progress](https://github.com/ansible-community/ara/pull/342) due for the upcoming release, a tasks page which will be analogous to /api/v1/tasks

We could probably imagine that there could be a "plays" and a "results" tab at some point in the future, each analogous to their respective API endpoint. It wasn't planned to be that way from the start but it evolved and grew organically as I learned about django and DRF in the process :p

Anyway, I guess what I am trying to say is that there isn't a model built around users and controllers but I can see why one might want to group playbooks under them. How would you see this work in practice ? What would the page look like ?

In terms of implementation at the API level, one option I see would be to split out the controller and executor out to their respective model tables (not unlike hosts) but that's a bit more complex than two simple varchar fields so it needs to be worth it :)

ara/cli/record.py
mocdaniel: Thanks for sharing some practical examples. The right fix from my perspective would be to 

* either go with the proposed change of introducing `recordkey` and refactor all the code which relates to this
* Rename the clashing `key` argument used for the client certificate's private key to something like `privkey` or `certkey` and refactor those parts of the codebase.

Anyways, this should probably not be part of this PR but raised as separate issue and handled as separate PR - I'd rollback my changes to `ara/cli/record` for this PR in the meantime.

mocdaniel:
> I think the question we could ask ourselves if whether it is fatal if the callback plugin is somehow unable to retrieve the hostname and the username. I haven't seen it happen but I suppose there could be a case where, somehow, we can't get the hostname properly and then because we [raise an exception](https://github.com/ansible-community/ara/blob/7f8b8367e0ec5083bab836422025f699780cb9f7/ara/plugins/callback/ara_default.py#L739-L741), that prevents the playbook from being recorded at all.
> 
> Looking at it from that angle, I'm not sure whether it should really be fatal since we could probably very well record 99% of the remainder of the playbook. We could make the fields nullable and default to None if we can't recover it instead ?
> 
> In the UI or CLI, we could decide whether to show anything at all if we don't have a value, write `n/a` or something.

I'm with you on that one - not being able to fetch a username/host should not be fatal, as there's still enough valuable information to record from a Playbook run. I personally like the idea of making the `executor` field nullable and displaying `n/a` in the Web UI in those rare cases.

Regarding the proposed change to the Web UI - I see your concerns, the different considerations and the work that plays into that, so my proposition would be to see if we feel the need for such a grouping in our use-cases once we start recording `executors` and come back to raise a PR if we do.

ara/cli/record.py
dmsimard: This is a good catch :)

It's a breaking change to rename the arguments but it's also a new feature that hasn't yet been released so I will let it go without a deprecation cycle dance -- it's only in the master branch:
- https://github.com/ansible-community/ara/pull/392
- https://github.com/ansible-community/ara/pull/403

Since I'm the one who chose the initial argument names, I volunteer to replace ``--key``, ``--cert`` and ``--ca`` by ``--ssl-key``, ``--ssl-cert`` and ``--ssl-ca`` in another PR if no one gets to it first.

FYI @devon-mar

dmsimard:
By the way, how did you come up with "executor" as the name and how strongly do you feel about it ?

Naming things is hard and I ponder whether ``user`` or something else would be a better fit.
I think user could be confusing because it could mean the remote user or perhaps the user that authenticated to send the data to the API.

I'll sleep on this one but curious to hear what you think :)

dmsimard:
> I'm with you on that one - not being able to fetch a username/host should not be fatal, as there's still enough valuable information to record from a Playbook run. I personally like the idea of making the executor field nullable and displaying n/a in the Web UI in those rare cases.

Ok, then we can handle exceptions by defaulting to None here and we can adjust the hostname one later.

mocdaniel:
> By the way, how did you come up with "executor" as the name and how strongly do you feel about it ?
> 
> Naming things is hard and I ponder whether `user` or something else would be a better fit. I think user could be confusing because it could mean the remote user or perhaps the user that authenticated to send the data to the API.
> 
> I'll sleep on this one but curious to hear what you think :)

The same thought process you had right now - `user` or `username`'s intent could be misunderstood in this context, and `executing_user` felt weird as the URL query argument would become `executing_user` then, too, instead of e.g. `executing-user` which feels more 'URL-like'.



> 

I'll commit the changes along with the rollback of the `--key` argument changes later today.

hille721:
> By the way, how did you come up with "executor" as the name and how strongly do you feel about it ?

@dmsimard good point, wanted to meantion the same. 

I'm really struggling with the wording "executor". What we will store here is the username under which the playbook is running. But this is in gerneral not equal with the "executor". 
You could run Ansible under root or a service account. Or in an execution environment inside a container, whre you will have a different username or maybe even no username. Further example is using AWX or Rundeck, also here we will just store just a system account (in case of Rundeck it is simple rundeck) but not the real executor, meaning the person who triggered the playbook.

mocdaniel:
This is true, I did not think about things like pipelines, AWX etc. issuing runs. What about `usercontext`? Apart from that one I'd need to think about the naming again...

ara/plugins/callback/ara_default.py
```diff
@@ -331,6 +332,10 @@ def v2_playbook_on_start(self, playbook):
         # Lookup the hostname for localhost if necessary
         self.localhost_hostname = self._get_localhost_hostname()
 
+        # TODO: create a function similar to _get_localhost_hostname with error handling (if needed)
+        # Lookup executing username
+        self.executor = pwd.getpwuid(os.getuid())[0]
```
hille721, line 337: out of curiosity, why you decided for this way to get the username. 
Here two function calls are needed, while there are possibilities with just one call, e.g. with the `getpass` library: `getpass.getuser( )`. 
The additional advantage here is also that this function first checks for commen env variables (`USERNAME`, `USER`, `LOGNAME`) and only if they are not set it will use the `pwd` module. 
The pwd module will directly do something similar as `getent passwd`. This is fine on systems with only local user entries. But can be more problematic on systems with e.g. LDAP, because it will increase the LDAP queries (if it is not cached) and will also not work if there any problems with LDAP.

In following example I disabled LDAP to simulate a LDAP error:
```
>>> import getpass
>>> getpass.getuser()
'hille'
>>> import os, pwd
>>> pwd.getpwuid(os.getuid())[0]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'getpwuid(): uid not found: 28242'
```

While `getpass.getuser()` is still working, `pwd` is not working anymore.

mocdaniel: Good catch! There's no specific reason *for* my decision, it just was the first (allegedly) reasonably robust implementation that came to mind. This judgement was apparently wrong, I'll go ahead and implement the lookup with `getpass.getuser()`.

mocdaniel:
Changes:
* Renamed `Playbook` model's field from `executor` to `usercontext`
* Set field `usercontext` to nullable and changed UI behaviour accordingly
* Fixed a small bug introduced by our changes that threw errors when issuing `ara playbook list`

hille721:
> We also thought about introducing a third Navbar entry to the Web UI called Users or Executors which would group recorded Playbook runs by executing user, but wanted to discuss necessity and approach with the maintainers first. 

Wouldn't it be much easier to introduce the `user/executor/usercontext/you name it` simple as a filter, similar to the controller filter (https://demo.recordsansible.org/?controller=fedora.novalocal)?
Then you are also able to quicktly get all playbooks for a specific user. 
I don't see a technical reason why for that the user must get a own model, @dmsimard .

@DanOPT , @mocdaniel : another thing what came to my mind. You both are also working for a German company, don't you have concerns about people who will complaining about privacy / data protection? I could image that this could happen here in my company. Thus maybe it would be also a good idea to have a setting to disable the user recording, to ensure this will be no blocking point to use Ara.
But this could be also done in a following PR

widhalmt:
> > @DanOPT , @mocdaniel : another thing what came to my mind. You both are also working for a German company, don't you have concerns about people who will complaining about privacy / data protection? I could image that this could happen here in my company. Thus maybe it would be also a good idea to have a setting to disable the user recording, to ensure this will be no blocking point to use Ara. But this could be also done in a following PR.
> 
> I don't really see this as a problem. In corporate environments you are subject to recording, monitoring and supervision already, especially when working on production infrastructure (the infamous "this incident will be reported" following a `sudo` without permission), and when hosting ARA at home to keep track of your personal Ansible runs this shouldn't matter, either.
> 
> @widhalmt what's your take on this? You've got much more experience regarding possible blocking points for software collecting data than I do. Right now, I'd go with @hille721 's opinion: This could be done in a following PR.

Same for me. I think it's a good point and there should be an issue created right after merging this PR. Usually you have a fixed team running Ansible code and it's usually not a problem to find a solution to track specific data of a small team of admins.

Making user name collection optional in the long run is a good idea, though. So people see, someone cares about privacy.

hille721:
> > Wouldn't it be much easier to introduce the `user/executor/usercontext/you name it` simple as a filter, similar to the controller filter (https://demo.recordsansible.org/?controller=fedora.novalocal)? Then you are also able to quicktly get all playbooks for a specific user. I don't see a technical reason why for that the user must get a own model, @dmsimard .
> 
> I'm a bit confused - isn't that exactly what we implemented? `usercontext` is just another charfield column in the `Playbook` table, just as `controller` is. The new UI 'tab' would've been another idea, but we moved away from that for now.

sorry my mistake, didn't checked in detail. Was just the false assumption that you haven't seen this great filter possibility. :)

mocdaniel:
Then I'd consider this PR done, from our POV. Feel free to merge at will and thanks for the great collaboration experience!

mocdaniel:
No hurries! This is totally understandable and the way it should be handled, anyways. Take your time. I just meant to say that we won't push new commits to this PR if not tasked to do so in my last comment.

dmsimard:
> > @DanOPT , @mocdaniel : another thing what came to my mind. You both are also working for a German company, don't you have concerns about people who will complaining about privacy / data protection? I could image that this could happen here in my company. Thus maybe it would be also a good idea to have a setting to disable the user recording, to ensure this will be no blocking point to use Ara. But this could be also done in a following PR.
> 
> I don't really see this as a problem. In corporate environments you are subject to recording, monitoring and supervision already, especially when working on production infrastructure (the infamous "this incident will be reported" following a `sudo` without permission), and when hosting ARA at home to keep track of your personal Ansible runs this shouldn't matter, either.
> 
> @widhalmt what's your take on this? You've got much more experience regarding possible blocking points for software collecting data than I do. Right now, I'd go with @hille721 's opinion: This could be done in a following PR.

Obligatory I am not a lawyer :) 

ara logs privileged information like ansible roles and playbook files, the playbook output, remote host facts, etc, but it doesn't have something like a privacy policy or something like GDPR compliance in mind because it's not a hosted service, it's an open source project you deploy yourself.

However, because there are use cases where it may not be preferred to record some information, ara provides ways to "ignore" various information like from the [docs](https://ara.readthedocs.io/en/latest/ansible-plugins-and-use-cases.html#ara-default-callback-recording-playbooks-from-ansible):
```
export ARA_IGNORED_FACTS=ansible_env,ansible_all_ipv4_addresses
export ARA_IGNORED_FILES=.ansible/tmp,vault.yaml,vault.yml
export ARA_IGNORED_ARGUMENTS=extra_vars,vault_password_files
```

I think it is appropriate and reasonable to provide a configuration boolean to disable collecting the controller hostname and user.
We should do it. It's not about compliance, it's about doing the right thing because I can see the use case for it.

Privacy is important and I've even gone to a certain length to [take out an unnecessary call to google](https://github.com/ansible-community/ara/commit/d7ecf9148755bbbe18ebeae2a43f79fc45555974) for a font :p

We can do it in another PR and do the controller hostname at the same time, let's try to land this first.

dmsimard:
I had time to look at this, it works, it's great :)

Screenshot from 2022-09-09 21-11-56

Screenshot from 2022-09-09 21-13-20

Looking at these ``n/a``, I think it would be better if we left them blank. If we picture a case where we don't save either the controller hostname or the user, it would be a lot of ``n/a`` :p

Otherwise, naming things is hard and I'm not sure about usercontext. I need to think about it.

I went through the exercise of finding out if ``user`` would be a bad idea. The CLI has a ``--username`` global parameter (for authenticating against the API) which could make an extra ``--user`` confusing but elsewhere I think it would be fine ?

If we consider that we're changing ``--key`` to ``--ssl-key`` because it clashed elsewhere and it's more straightforward that way, we could consider prefixing ``--username`` as ``--api-username`` (this one with a deprecation cycle).

The UI doesn't have a concept of user right now, the authentication is managed via django's basicauth or a proxy like nginx in front.

I don't have a better suggestion right now, are there strong objections to calling it user ?
Of course, I apologize if it means ``s/usercontext/user/`` everywhere.

Another thought: this records the username that ran the playbook but I figure it would probably be valuable to record the username that the user connected as, as well -- that wouldn't go in the playbook table though, it would go in the hosts table.

Out of scope for 1.6.0 though :p

ara/api/migrations/0012_playbook_usercontext.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.15 on 2022-09-05 15:23
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0011_play_label_name_max_length'),
```
dmsimard, line 9: Just a note so we don't forget about it -- this might change and needs rebasing depending on which one merges first, either this one or https://github.com/ansible-community/ara/pull/342 since it introduces a database migration too.

mocdaniel: Sure thing, just let us know once this PR is about to get merged.

mocdaniel:
> I had time to look at this, it works, it's great :)

Thank you! 

> Looking at these `n/a`, I think it would be better if we left them blank. If we picture a case where we don't save either the controller hostname or the user, it would be a lot of `n/a` :p

This is true, n/a's might stack up to point where the UI feels bloated yet empty at the same time. I'll adjust the templates.

> If we consider that we're changing `--key` to `--ssl-key` because it clashed elsewhere and it's more straightforward that way, we could consider prefixing `--username` as `--api-username` (this one with a deprecation cycle).
> I don't have a better suggestion right now, are there strong objections to calling it user ? Of course, I apologize if it means `s/usercontext/user/` everywhere.

The similarity between `--username` and `--user` is a head scratcher, yes. No strong objections though, so I'll go ahead and change `usercontext` to `user`.

ara/ui/forms.py
```diff
@@ -9,6 +9,7 @@
 class PlaybookSearchForm(forms.Form):
     ansible_version = forms.CharField(label="Ansible version", max_length=255, required=False)
     controller = forms.CharField(label="Playbook controller", max_length=255, required=False)
+    user = forms.CharField(label="User context", max_length=255, required=False)
```
dmsimard, line 12:
```suggestion
    user = forms.CharField(label="Playbook user", max_length=255, required=False)
```

ara/ui/templates/index.html
```diff
@@ -36,6 +36,10 @@
                 <label for="controller" {% if request.GET.controller %}style="font-weight:bold;"{% endif %}>Ansible controller</label>
                 <input type="text" class="form-control" id="controller" name="controller" placeholder="ex: host.domain.tld (or) host" value="{{ search_form.controller.value | default_if_none:'' }}">
               </div>
+              <div class="form-group col-xl-5">
+                <label for="user" {% if request.GET.user %}style="font-weight:bold;"{% endif %}>User context</label>
```
dmsimard, line 40:
```suggestion
                <label for="user" {% if request.GET.user %}style="font-weight:bold;"{% endif %}>User</label>
```

ara/ui/templates/index.html
```diff
@@ -112,6 +116,7 @@
             <th class="text-center" title="Number of results recorded from the playbook">Results</th>
             <th class="text-center" title="Version of Ansible that ran the playbook">Ansible</th>
             <th class="text-center" title="Host that ran the playbook">Controller</th>
+            <th class="text-center" title="User context of playbook run">User context</th>
```
dmsimard, line 119:
```suggestion
            <th class="text-center" title="User that ran the playbook">User</th>
```

ara/ui/templates/index.html
```diff
@@ -146,11 +151,14 @@
             </td>
             <td class="text-center">
               {% if not static_generation %}
-                <a href="{% url 'ui:index' %}?controller={{ playbook.controller }}">{{ playbook.controller }}</a>
+                <a href="{% url 'ui:index' %}?controller={{ playbook.controller }}">{{ playbook.controller  }}</a>
```
dmsimard, line 154:
```suggestion
                <a href="{% url 'ui:index' %}?controller={{ playbook.controller }}">{{ playbook.controller }}</a>
```

Looks like an unintended extra whitespace here

dmsimard:
@mocdaniel left a couple suggestions, after that it's good to go. We'll land this before the new tasks page so I'll be the one rebasing the SQL migration :)

ara/ui/forms.py
mocdaniel: Fixed.

ara/ui/templates/index.html
mocdaniel: Fixed.

ara/ui/templates/index.html
mocdaniel: Fixed.

ara/ui/templates/index.html
mocdaniel: Good catch :D Fixed.

mocdaniel:
> @mocdaniel left a couple suggestions, after that it's good to go. We'll land this before the new tasks page so I'll be the one rebasing the SQL migration :)

Implemented the suggestions, thanks again for the pleasant contributor experience! LGTM :)

dmsimard:
> thanks again for the pleasant contributor experience!

Likewise, it has been pleasant to collaborate with you :tada:

I'll wait for the CI to chime in and will merge afterwards.

dmsimard:
There you go!

I've created an issue so we don't forget the optional toggle before the release: https://github.com/ansible-community/ara/issues/425

dmsimard:
By the way, something I didn't notice during the review is that the tooltip for the user search is None:
Screenshot from 2022-09-17 18-24-40

There should be an example like the other fields. I can eventually take care of it but feel free to send a patch :p

dmsimard:
Fixed here: https://github.com/ansible-community/ara/pull/426

dmsimard:
FYI, I've sent a PR to make recording controller hostnames and users optional: https://github.com/ansible-community/ara/pull/438

It'll land in the upcoming release of 1.6.0.

PR #419, opened 2022-09-10
Disable InsecureRequestWarning for insecure requests

### LINKED ISSUE #356: Add option for trusted CA when using HTTPS
## What is the idea ?
There is an ARA_API_INSECURE option which lets one ignore SSL errors, but then you get a python warning message at runtime. Would be nice for an option like ARA_API_CA which can point to a file that would be used by the requests library when making HTTPS requests.

gaby:
The purpose of this PR is to disable the requests warning when using `--insecure` or `ARA_API_INSECURE`

Fixes #356

### DIFF
--- a/ara/clients/http.py
+++ b/ara/clients/http.py
@@ -7,6 +7,7 @@ import weakref
 
 import pbr.version
 import requests
+import urllib3
 
 from ara.clients.utils import active_client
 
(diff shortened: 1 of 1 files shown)


dmsimard:
Works for me. Thanks !

PR #421, opened 2022-09-11
Auto format codebase to py36+

### LINKED ISSUE #420: Drop support for Python EOL versions
## What is the idea ?
Looking at the documentation and code base, it looks like `ara` still supports or tries to support Python 3.5 and 3.6. These versions of Python are EOL since 2020 for 3.5 and 2021 for 3.6

Are there plans to address this?

Source: https://devguide.python.org/versions/

gaby:
* This PR auto-formats all the code-base to py36+ using `pyupgrade`: https://github.com/asottile/pyupgrade

Command: 
```bash
git clone git@github.com:ansible-community/ara.git
pyupgrade --keep-percent-format --py36-plus *.py */*/*/*/*/*.py
```

* Sorted all imports using `isort`: https://github.com/PyCQA/isort

Command: 
```bash
git clone git@github.com:ansible-community/ara.git
isort .
```

Partially Fixes #420

### DIFF
--- a/ara/clients/http.py
+++ b/ara/clients/http.py
@@ -14,7 +14,7 @@ from ara.clients.utils import active_client
 CLIENT_VERSION = pbr.version.VersionInfo("ara").release_string()
 
 
-class HttpClient(object):
+class HttpClient:
     def __init__(self, endpoint="http://127.0.0.1:8000", auth=None, cert=None, timeout=30, verify=True):
         self.log = logging.getLogger(__name__)
 
@@ -64,7 +64,7 @@ class HttpClient(object):
         return self._request("delete", url)
 
 
-class AraHttpClient(object):
+class AraHttpClient:
     def __init__(self, endpoint="http://127.0.0.1:8000", auth=None, cert=None, timeout=30, verify=True):
         self.log = logging.getLogger(__name__)
         self.endpoint = endpoint
@@ -85,12 +85,12 @@ class AraHttpClient(object):
             response = func(url, **kwargs)
 
         if response.status_code >= 500:
-            self.log.error("Failed to {method} on {url}: {content}".format(method=method, url=url, content=kwargs))
+            self.log.error(f"Failed to {method} on {url}: {kwargs}")
 
-        self.log.debug("HTTP {status}: {method} on {url}".format(status=response.status_code, method=method, url=url))
+        self.log.debug(f"HTTP {response.status_code}: {method} on {url}")
 
         if response.status_code not in [200, 201, 204]:
-            self.log.error("Failed to {method} on {url}: {content}".format(method=method, url=url, content=kwargs))
+            self.log.error(f"Failed to {method} on {url}: {kwargs}")
 
         if response.status_code == 204:
             return response
--- a/ara/plugins/lookup/ara_api.py
+++ b/ara/plugins/lookup/ara_api.py
@@ -1,13 +1,11 @@
 # Copyright (c) 2022 The ARA Records Ansible authors
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-from __future__ import absolute_import, division, print_function
 
 from ansible.plugins.lookup import LookupBase
 
 from ara.clients import utils as client_utils
 
-__metaclass__ = type
 
 DOCUMENTATION = """
     lookup: ara_api
@@ -37,7 +35,7 @@ RETURN = """
 
 class LookupModule(LookupBase):
     def __init__(self, *args, **kwargs):
-        super(LookupModule, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
         self.client = client_utils.active_client()
 
     def run(self, terms, variables, **kwargs):
--- a/ara/plugins/lookup/ara_api.py
+++ b/ara/plugins/lookup/ara_api.py
@@ -6,7 +6,6 @@ from ansible.plugins.lookup import LookupBase
 
 from ara.clients import utils as client_utils
 
-
 DOCUMENTATION = """
     lookup: ara_api
     author: David Moreau-Simard (@dmsimard)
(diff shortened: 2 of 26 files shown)


ara/clients/http.py
```diff
@@ -85,12 +85,12 @@ def _request(self, method, url, **kwargs):
...
         if response.status_code >= 500:
-            self.log.error("Failed to {method} on {url}: {content}".format(method=method, url=url, content=kwargs))
+            self.log.error(f"Failed to {method} on {url}: {kwargs}")
 
-        self.log.debug("HTTP {status}: {method} on {url}".format(status=response.status_code, method=method, url=url))
+        self.log.debug(f"HTTP {response.status_code}: {method} on {url}")
 
         if response.status_code not in [200, 201, 204]:
-            self.log.error("Failed to {method} on {url}: {content}".format(method=method, url=url, content=kwargs))
+            self.log.error(f"Failed to {method} on {url}: {kwargs}")
```
dmsimard, line 93: Yay, f-strings :tada:

ara/plugins/lookup/ara_api.py
```diff
@@ -1,14 +1,11 @@
...
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-from __future__ import absolute_import, division, print_function
 
 from ansible.plugins.lookup import LookupBase
 
 from ara.clients import utils as client_utils
 
-__metaclass__ = type
-
```
dmsimard, line 11: These need to stay.

I don't make the rules: they're part of the Ansible module boilerplate i.e, https://github.com/ansible/ansible/blob/79f67ed56116be11b1c992fade04acf06d9208d1/lib/ansible/plugins/lookup/dict.py#L4-L5

ara/setup/action_plugins.py
```diff
@@ -1,7 +1,6 @@
 # Copyright (c) 2022 The ARA Records Ansible authors
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-from __future__ import print_function
```
dmsimard, line 4: Same thing as https://github.com/ansible-community/ara/pull/421/files#r967722815

ara/setup/ansible.py
```diff
@@ -1,7 +1,6 @@
 # Copyright (c) 2022 The ARA Records Ansible authors
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
-from __future__ import print_function
```
dmsimard, line 4: Same thing as https://github.com/ansible-community/ara/pull/421/files#r967722815

dmsimard:
Thanks for this ! It largely looks good to me except the bits in the Ansible plugins -- I didn't figure it was necessary to leave a comment in each of the case.

ara/clients/http.py
hille721: f-strings shouldn't be used in log module, instead the paramaters should be passed as arguments.

E.g.

``` python
log.error("Hello %s", world)
```

While f-strings will always be evaluated, this will only be evaluated if the log level is above the loggers log level.

See https://docs.python.org/3/howto/logging.html#logging-variable-data 
or https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/

By using pylint you would also get a warning (https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html)

PS: also the old code with `{}.format` should not be used in log module. But if we change it now, we can make it directly perfect.

gaby: @hille721 Thanks for the information. I did some google'ing and found this one too. 

https://docs.python.org/3/howto/logging.html#optimization

I will make the changes to use % for log messages, and also run pylint

gaby: I went ahead and opened a bug report on pyupgrade to address this. 

Ref https://github.com/asottile/pyupgrade/issues/702

gaby: Pyupgrade developer immediately closed by issue stating "I was wrong and should look closer".

hille721: maybe because it is not really related to pyupgrade. The behavior of the logging module didn't changed.
I guess pyupgrade is simple replacing str.format strings with f-strings. 
But yeah his answer is really harsh.

asottile: indeed -- and if you searched the tracker before making an issue you would have found a duplicate -- https://github.com/asottile/pyupgrade/issues?q=is%3Aissue+logging+is%3Aclosed -- garbage in garbage out -- technically pyupgrade _improved_ your performance here but it was incorrect to begin with

dmsimard: @asottile there was no need to be mean about it but understood, thanks.

asottile: not sure where you're interpreting "mean" -- the issue was invalid, a duplicate, no code sample, no version information, and therefore closed

gaby:
I plan on resolving all the comments tonight.

gaby:
@dmsimard @hille721 All comments have been resolved. Let me know if you need me to change anything else. 

As stated in #424 there some linter improvements that can be done. I noticed when running `black .` on my branch I get this: 

image

gaby:
Should I apply the changes from `black` to this PR ?

hille721:
> As stated in #424 there some linter improvements that can be done. I noticed when running `black .` on my branch I get this:

you should not run `black  .`. We are having a specifig black config `.black.toml`, thus you have to run `black --config .black.toml` or even better calling `tox -e linters`. This will also run autoamtically on every commit.

gaby:
> > As stated in #424 there some linter improvements that can be done. I noticed when running `black .` on my branch I get this:
> 
> you should not run `black .`. We are having a specifig black config `.black.toml`, thus you have to run `black --config .black.toml` or even better calling `tox -e linters`. This will also run autoamtically on every commit.

Ahhh, I see. I was assuming that black would use `.black.toml` by default like other cli's do. After running it with ` black --config .black.toml .` I only get 3 files with changes which are the same ones you committed in #424

hille721:
exactly, not that worse :)

dmsimard:
@gaby could you rebase on top of latest master ? I think it's good to go afterwards.

gaby:
@dmsimard Done

dmsimard:
Looks good to me, the code feels a bit cleaner now I guess ¯\\_(ツ)_/¯

Thanks !

PR #432, opened 2022-09-19
filter link on user field

### LINKED ISSUE #431: New user column does not provide a quick filter link
## What is the issue ?
image

Clicking on the username should also filter for this particular user, similar as we have it with the other columns:

hille721:
* link on user column to quickly filter on it
* added title for this column
* also added missing title for ansible version and controller
* left aligned user column

image

(fixes #431)

### DIFF
--- a/ara/ui/templates/index.html
+++ b/ara/ui/templates/index.html
@@ -116,7 +116,7 @@
             <th class="text-center" title="Number of results recorded from the playbook">Results</th>
             <th class="text-center" title="Version of Ansible that ran the playbook">Ansible</th>
             <th class="text-center" title="Host that ran the playbook">Controller</th>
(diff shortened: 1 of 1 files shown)


hille721:
> Based on your comment, should we left align the controller too ?

Also thought about  if I should change this in this PR as well. But there the central alignment doesn't really bother me because the controller / the controller length is often the same, at least at those instances where I'm checking the look and feel (my companies instances, test instance, demo.recordsansible.org).

But yeah if the lenght of the controller differs a lot, e.g. in an ansible-pull setup with non standardized hostnames, this could look better as well. 

What is your feeling?

dmsimard:
No strong feeling either way. Let's leave it like this for now.

Issue #434, opened 2022-09-27
ara should probably record the version of the callback plugin and API server


dmsimard:
## What is the idea ?

The version of the http client is sent as a user-agent ([here](https://github.com/ansible-community/ara/blob/ed20abc682255bb9f54df0a390147fee857a7a56/ara/clients/http.py#L27)) and it shows up in server logs kind of like this:
``"GET /api/v1/playbooks?order=-started&limit=50 HTTP/1.1" 200 66635 "-" "ara-http-client_1.5.9.dev51"``

The version isn't saved in the database like the Ansible version is, though.
It may be useful to save it as there may be disparities with different behaviors across versions.

I think ideally everyone has the same version everywhere but mismatches are bound to happen.

It would be inexpensive to implement even if just for troubleshooting and debugging purposes, just extra fields next to Ansible version: https://github.com/ansible-community/ara/blob/ed20abc682255bb9f54df0a390147fee857a7a56/ara/api/models.py#L86

PR #436, opened 2022-10-01
Add support for recording versions of ara, fix async task recording with UUIDs

dmsimard:
Related issue: https://github.com/ansible-community/ara/issues/434

Recording versions of python, the ara client and server that were involved may
prove helpful for comparing recordings across versions and
troubleshooting issues.

This commit includes a database migration for adding the
"client_version", "server_version" and "python_version" fields
to the playbook model, API and CLI commands.

In a typical scenario, the callback plugin supplies it's version as
client_version and the server_version field is automatically populated
by the server.

The UI implementation is limited for this first iteration:
- It is possible to search for ?client_version, ?server_version and
  ?python_version parameters by manually specifying them in the playbook
  index URL: there are no search form fields for them
- The versions are only shown in the playbook report card at the top of the
  page.

Drive-by improvements:
- Fix phrasing for --user help on the playbook CLI commands
- Set the version of ara once under ara.setup and use that everywhere
  instead of calling pbr.version every time
- Re-order columns in the playbook report card such that item counts are
  further to the right as they are less relevant in that context


dmsimard:
I've spent quite a bit of time on this -- tested it locally in addition to the integration tests in this PR and in [ara-collection](https://github.com/ansible-community/ara-collection/pull/60).

After a couple reviews and iterations, it's good enough to land.

I came across unrelated things to improve or fix and this PR has already gotten large enough, I'll open another one.

Issue #425, opened 2022-09-12
Make controller hostname and username recording optional


dmsimard:
## What is the idea ?

Recording the controller hostname and username has privacy implications and users may want to opt out of recording them.

There should be callback settings (somewhat like ``IGNORED_FILES``) that disables recording for them and sets their value to None instead.

amilcarlucas:
This is nice. But using ara 1.5.8 I do not see the "user that run the playbook" nowhere in the web interface. Where is it?

hille721:
new feature which will come with the 1.6.0 release.
There is already a release candidate available to test: https://pypi.org/project/ara/1.6.0.0rc1

hille721:
@dmsimard I think this issue can be closed because #438 is already merged

dmsimard:
Yes indeed, thanks @hille721.

Sorry about the delay in releasing 1.6.0. It is overdue.

PR #438, opened 2022-10-08
callback: Make user and controller recording optional

dmsimard:
While ara will record the user and controller by default, it can now be made optional with new settings:

- ARA_RECORD_USER (or record_user in ansible.cfg)
- ARA_RECORD_CONTROLLER (or record_controller in ansible.cfg)

This brings their recording in line with other components that can be opted-out like host facts, files and command line arguments.


dmsimard:
Related issue: https://github.com/ansible-community/ara/issues/425

PR #439, opened 2022-10-08
ui: re-order table columns and improve consistency

dmsimard:
This pushes information that are less relevant further away to the right such that the more relevant things are closer to the left.

### DIFF
--- a/ara/ui/static/css/ara.css
+++ b/ara/ui/static/css/ara.css
@@ -37,14 +37,56 @@
 }
 
 .custom-checkbox .custom-control-input:checked ~ .custom-control-label::before {
-  background-color: #b4bcc2;
+    background-color: #b4bcc2;
 }
 
 .custom-checkbox .custom-control-input:active ~ .custom-control-label::before {
-  background-color: #ffffff;
+    background-color: #ffffff;
 }
 
 .tooltip {
-  pointer-events: none;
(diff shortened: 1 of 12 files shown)


dmsimard:
@hille721 what do you think about this one ? The column orders felt inconsistent across playbooks, hosts and tasks and I thought this made sense.

For comparison (before and after):

### Playbooks

Selection_039

Selection_041

### Tasks

Selection_040

Selection_042

dmsimard:
This has become a bit more extensive than I had originally intended but it bothered me :sweat_smile:

Comparison below (play both videos at the same time to easily view the difference)

### Before

https://user-images.githubusercontent.com/1291204/194768358-9c55f013-d3ed-44fc-ad92-a2828dea337a.mp4

### After

https://user-images.githubusercontent.com/1291204/194768372-030a2467-fe89-4004-9999-c7741d1b0497.mp4

dmsimard:
I've brought in changes from https://github.com/ansible-community/ara/pull/443 because they didn't rebase cleanly on top of this and it was a pain to test.

hille721:
> @hille721 what do you think about this one ? The column orders felt inconsistent across playbooks, hosts and tasks and I thought this made sense.

Agree, makes sense.

dmsimard:
Let's ship it then !

Issue #136, opened 2020-05-20
server error in ara-web console


hritz220:
I'm trying to use distributed sqlite, but when I copy the database in /var/www/log/ara-report/ansible.sqlite , I'm unable to see data in the ara-web console.
Here is the set up. Installed ara with API server dependencies
`python3 -m pip install --user ansible "ara[server]"`
Updated settings.yml
`DISTRIBUTED_SQLITE: true`
Started the server
`ara-manage runserver localhost:80`
generated and copied the ansible.sqlite from another machine to the location /var/www/log/ara-report/ansible.sqlite 
when I hit the URL: http://localhost/ara-report

```

root@ritik-Latitude-E7440:/home/ritik# ara-manage runserver localhost:80
[ara] Using settings file: /root/.ara/server/settings.yaml
[ara] Using settings file: /root/.ara/server/settings.yaml
2020-05-15 16:53:00,734 INFO django.utils.autoreload: Watching for file changes with StatReloader
Performing system checks...
System check identified no issues (0 silenced).
You have 23 unapplied migration(s). Your project may not work properly until you apply the migrations for app(s): admin, api, auth, contenttypes, db, sessions.
Run 'python manage.py migrate' to apply them.
May 15, 2020 - 16:53:01
Django version 2.2.12, using settings 'ara.server.settings'
Starting development server at http://localhost:80/
Quit the server with CONTROL-C.
2020-05-15 16:53:10,429 ERROR django.request: Internal Server Error: /ara-report/
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.6/dist-packages/django/db/backends/sqlite3/base.py", line 383, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.OperationalError: no such table: playbooks
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/dist-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/dist-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
[...]

PR #442, opened 2022-10-09
ui: Add a simple index for the distributed sqlite backend

dmsimard:
The index when using the distributed sqlite backend was a 404 not found which isn't very helpful.

Add a simple (but useful) index that contains a list of databases and we can improve it later.

### DIFF
--- a/ara/server/wsgi.py
+++ b/ara/server/wsgi.py
@@ -52,6 +52,12 @@ def distributed_sqlite(environ, start_response):
     if path_info.startswith(settings.STATIC_URL) or path_info == "/healthcheck/":
         return default_application(environ, start_response)
 
+    # The root of the application should be served by the regular app to show
+    # a distributed database index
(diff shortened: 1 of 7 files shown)


dmsimard:
This wasn't planned for 1.6.0 but I happened to need it and it has been [asked about in the past](https://github.com/ansible-community/ara/issues/136#issuecomment-640179155).

It looks like this:

Screenshot from 2022-10-08 21-22-14

apollo13:
This looks great but `lru_cache` is probably not a good idea. What about instead of a single view (that is hard to scale) have a directory "browser" like a normal webserver index page?

dmsimard:
After discussing with you a bit, I realized that my understanding of how LRU cache worked was wrong so I'll take it out for now and will figure something out later.

Thanks !

dmsimard:
This will work for now, let's revisit it later to make sure it's faster and scales better.

PR #443, opened 2022-10-09
Misc UI fixes

dmsimard:
- ui: Truncate table fields which can be exceedingly long
- ui: Fix host links to use names instead of id

### DIFF
--- a/ara/ui/templates/index.html
+++ b/ara/ui/templates/index.html
@@ -178,7 +178,7 @@
             <td title="{{ playbook.path }}">
               {% if not static_generation %}
                 {% if playbook.name is not None %}
-                  <a href="{% url 'ui:index' %}?name={{ playbook.name }}">{{ playbook.name }}</a>
+                  <a href="{% url 'ui:index' %}?name={{ playbook.name }}">{{ playbook.name | truncatechars:50 }}</a>
                 {% else %}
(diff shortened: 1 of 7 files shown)


dmsimard:
I rebased these changes on top of https://github.com/ansible-community/ara/pull/439 instead.

hille721:
@dmsimard : not sure if the change in the links from id to name is really good. The host name filter is only a contain filter, not equal filter. Thus wil lead to a lot of false positive results. What is wrong with having the id in the linked URL which easily gives exactly the results which are needed.

dmsimard:
@hille721 The issue I found and tackled this way is that filtering by ID worked but didn't populate the host name search field -- so while there was a query, there was nothing in the search form at all.

Same thing for task names and I realize that it could lead to false positives.

Should we add the ID fields back ? ¯\\_(ツ)_/¯

hille721:
@dmsimard : instead of filtering a playbook for a certain host (`/playbooks/playbookId.html?host=hostId`) we should directly go to `/hosts/hostId.html`, then we also don't have that problem what you mentioned. 

But I'm confused because in the current master code it is already like that: 

https://github.com/ansible-community/ara/blob/master/ara/ui/templates/partials/playbook_hosts.html#L40

Thus for me there is no change needed.

PR #446, opened 2022-10-10
improved host linking

hille721:
Link to `/hosts/hostID.html` instead of searching for host name on playbook site. As discussed searching by host name will lead to false positives and also searching by host id makes not really sense, because we can directly go to the host site itself.

The link in the playbook hosts card will also directly point to the facts section. Because I saw that you will move the fact section to the bottom and this would ensure that this section is still fastly available.


dmsimard:
Yeah this is probably fine.

I suppose we have a similar issue with task names but we don't (yet?) have a dedicated page for tasks like we do for hosts and playbooks.

hille721:
> I suppose we have a similar issue with task names but we don't (yet?) have a dedicated page for tasks like we do for hosts and playbooks.

Yes, but I think the issue is not that bad for tasks. I assume less false positive results here because task names will be more various.
But yeah generally it could make sense to have a dedicated task page as well.

flowerysong:
I'm not a fan of the link being to a specific fragment; if you think the ordering of the page emphasizes the wrong information that should be addressed by changing the page, not by unexpectedly linking the user to the bottom of the page.

dmsimard:
@flowerysong I may have taken the wrong route when I recently suggested we put host facts at the bottom (UX is hard).

My own experience as a user is that I would, more often than not, need to scroll down to the bottom or collapse the facts card to see the task results. Sure, sometimes the facts are what I'm looking for and it's great that they're available.

I think the issue I have with it is that it pushes too much of the content down.

Perhaps we should limit it's max height a bit like we do for the files card in the playbook page:
Peek 2022-10-10 21-25

Instead of what we have now:
Peek 2022-10-10 21-26

I would have less of an issue removing the ``#headers`` links if we don't push the content down so much.

dmsimard:
I tested to see what it would look like, it would look like this:
Peek 2022-10-10 22-02

hille721:
basically I would prefer to have the facts at the bottom with more space. With this less space it will be really hard. There are facts which will be not completely visible without scrolling. 

I routed the link to facts directly because there is already the report link at the beginning which routes to the page. (makes no sense for me having the same link directly close to each other). But well also fine if we add a further column "facts" which is directly linking to this section.

dmsimard:
@hille721 @flowerysong 

I've been using the release candidate for a while and I was pondering whether we should revert both this change and the one that moved the facts below the task results.

I justified the change on [personal experience](https://github.com/ansible-community/ara/commit/f924fdeafdcfc79a2d8f8281b3e06b356c48dbfe) but I am not convinced after all.

I guess my question is: was it better before ?

If you want to compare with a current installation, perhaps, https://demo.recordsansible.org is up to date with the latest master branch.

UX is hard. I welcome any ideas and contributions :p

dmsimard:
In absence of feedback or strong opinions one way or another, I prefer reverting for the time being.
We keep what we had before without needlessly changing it while users have had time to get accustomed to the current implementation.

In other words: I am not confident enough about this change and feel stronger towards keeping host facts above the task results.

I think part of the solution in improving the UX of the various pages is to add navigation breadcrumbs at the top or something along those lines so we know there are more headers and items below, there just isn't enough height to display everything.

I will create an issue for that.

PR #448, opened 2022-10-11
1.6.0rc1 fixes

dmsimard:
I'll keep this PR opened for a while, we can put the fixes we find for rc1 here.

- ui: center the user cell in the playbook list table

### DIFF
--- a/ara/ui/templates/host.html
+++ b/ara/ui/templates/host.html
@@ -8,11 +8,11 @@
   </div>
   <br/>
   <div>
-    {% include "partials/host_results.html" %}
+    {% include "partials/host_facts.html" %}
   </div>
   <br/>
   <div>
-    {% include "partials/host_facts.html" %}
+    {% include "partials/host_results.html" %}
   </div>
   <br/>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


hille721:
The user cell left alignment was intentionally, not accidentally :) 

see: https://github.com/ansible-community/ara/pull/432

dmsimard:
> The user cell left alignment was intentionally, not accidentally :)
> 
> see comment [#431 (comment)](https://github.com/ansible-community/ara/issues/431#issuecomment-1251008718) and pr #432

Oh, I forgot. Let's keep it like that for now then.

UX is hard.

dmsimard:
@hille721 I'm re-opening this but will change the fix -- the table header should be aligned with the cells but right now it's centered:

Selection_060

dmsimard:
I've reverted the change in which we moved host facts below the task results to keep what we had before for now.

There are other ways to improve navigation and UX, one of which I've created an issue about: https://github.com/ansible-community/ara/issues/463

dmsimard:
Thanks for the :+1: @hille721, let's ship it

Issue #428, opened 2022-09-19
Option to change the application root path


mathishaloy:
## Option to change the application root path
In the previous version (0.x) it was possible to change the application root path (/) to other path like (/ara)
This option is not present in the latest documentation, is there a way today to change the application root path ?

Thanks !

dmsimard:
Hi @mathishaloy and thanks for the issue.

We recommend setting the API server at the root of a subdomain because it's simpler that way.

If you really want to put it under a subdirectory, there's a workaround suggested in this other issue: https://github.com/ansible-community/ara/issues/121

mathishaloy:
Hi @dmsimard, thanks for your quick response. Unfortunately I'm using the docker image, do you have also a workaround for this type of installation ?

dmsimard:
The workaround is the same: put an nginx in front of the container which takes care of modifying URLs to be under a subdirectory.

rptaylor:
I think this is addressed by https://ara.readthedocs.io/en/latest/api-configuration.html#ara-base-dir  from https://github.com/ansible-community/ara/pull/452  right?

PR #452, opened 2022-11-02
Allow ARA to be deployed on a "subpath".

apollo13:
By default ARA listens on `localhost:8000/`. Using BASE_PATH this can be changed to ie `/ara/` to have ARA listen on `localhost:8000/ara`. This is not really useful during development but can be helpful when a reverse proxy is in front of ARA.

@dmsimard I realize this is rather late in the 1.6 cycle but I'd love it to be in there.


dmsimard:
@apollo13 people have asked about this before (most recently [here](https://github.com/ansible-community/ara/issues/428)) but I didn't have good advice -- it was pretty tricky to do with Flask back in the day. I'm happy that it actually turns out to be simple with django.

I'm very late in releasing 1.6.0 due to various circumstances, we can probably include this change.

I will test and review it soon.

apollo13:
Mhm something is wrong here still, looking into it :)

apollo13:
I think it works okay now, currently testing against our AWX setup :)

dmsimard:
Hi @apollo13 and thanks for the PR.

I've tested this successfully using ``gunicorn --access-logfile - --bind 127.0.0.1:8000 ara.server.wsgi`` but it doesn't seem to work with ``ara-manage runserver``.

Is it expected to ?

Do we need to add something in https://github.com/ansible-community/ara/blob/master/ara/server/__main__.py or elsewhere ?

apollo13:
@dmsimard What exactly did you test? I just set `BASE_PATH` to `/test` and tested it with and without `DISTRIBUTED_SQLITE`. Please note that `/` will 404 and only `/test/` will give you something meaningful. Are you sure that `ara-manage` did run against the correct ara version?

dmsimard:
@apollo13 I checked out your branch, modified settings.yaml to have ``BASE_PATH`` to ``/test``:

Screenshot from 2022-11-20 12-06-53

Whether or not there is a trailing slash doesn't appear to make a difference but I could be missing something obvious.

apollo13:
image
This is what it looks like here; I do have a vastly newer Django -- maybe that makes a diff?

dmsimard:
Ohh, I was unexpectedly running off of an old tox venv which still had Django 2.2.

It works with Django 3.2.

My bad :facepalm:

PR #458, opened 2022-11-07
callback: Make task content recording optional

tomsiewert:
ara records the task content by default which could contain secrets. This can now be made optional with a new setting:

- ARA_RECORD_TASK_CONTENT (or record_task_content in ansible.cfg)

Signed-off-by: Tom Siewert <tom@siewert.io>

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -238,6 +238,17 @@ options:
     ini:
       - section: ara
         key: record_user
+  record_task_content:
+    description:
+      - Whether ara should record the content of a task
+      - Defaults to true, set to false for privacy or other use cases
+    type: boolean
+    default: true
+    env:
+      - name: ARA_RECORD_TASK_CONTENT
+    ini:
+      - section: ara
+        key: record_task_content
 """
 
 # Task modules for which ara should save host facts
@@ -308,6 +319,7 @@ class CallbackModule(CallbackBase):
         self.localhost_as_hostname_format = self.get_option("localhost_as_hostname_format")
         self.record_controller = self.get_option("record_controller")
         self.record_user = self.get_option("record_user")
+        self.record_task_content = self.get_option("record_task_content")
 
         # The intent for the ignored_files default value is to ignore the ansible local tmpdir but the path
         # can be changed by the user's configuration so retrieve that and use it instead.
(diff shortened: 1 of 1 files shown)


dmsimard:
Hi @tomsiewert and thanks for the PR.

I am curious, could you explain your use case for this ?

Generally you would want to set ``no_log: true`` on tasks such that they are not printed in the console stdout either.
ara won't pick up the content of tasks when no_log is used.

tomsiewert:
I agree that `no_log: true` would be the way to solve that.

> I am curious, could you explain your use case for this ?

Sure. There are a few reasons for us to disable the task content recording:

1. The result content is not required for us to be recorded by ara at all. We are more interested in what playbook is being executed by user X, how long it took and which hosts were included.
2. We generate ~180k-250k task results per day. The performance is already a bit decreased and pushing the result content would probably decrease the performance even more.
3. We have >120 roles and ~200 playbooks in our Ansible collection. We could either disable ara and look into every role for possible secret leaks or disable the task content recording at all and fix each role step by step to ensure a partial recording.

dmsimard:
> I agree that `no_log: true` would be the way to solve that.
> 
> > I am curious, could you explain your use case for this ?
> 
> Sure. There are a few reasons for us to disable the task content recording:
> 
>     1. The result content is not required for us to be recorded by ara at all. We are more interested in what playbook is being executed by user X, how long it took and which hosts were included.
> 
>     2. We generate ~180k-250k task results per day. The performance is already a bit decreased and pushing the result content would probably decrease the performance even more.
> 
>     3. We have >120 roles and ~200 playbooks in our Ansible collection. We could either disable ara and look into every role for possible secret leaks or disable the task content recording (temporarily) and fix each role step by step to ensure a partial recording.
> 
>     4. Human make mistakes. In case someone forgets to enable `no_log` on a task a secret could be leaked.

That makes sense, thanks !

I am trying to ship ara 1.6.0 and this won't make the cut but I will leave the PR open so I can review, test and merge it later.

Your patience is appreciated.

dmsimard:
Hi @tomsiewert and sorry it's taken so long.

There is nothing wrong with the code but it unfortunately ran into a merge conflict due to callback changes in https://github.com/ansible-community/ara/pull/496.

I'm not the kind to ask you to rebase after so long so I've taken upon myself to do it here: https://github.com/ansible-community/ara/pull/504

It works well and I've tested it locally.
I will close this PR in favor of #504 and will eventually follow up with a patch to add some testing coverage to the feature before 1.7.0 releases.

Thanks for your patience and contribution.

Issue #171, opened 2020-09-04
Measuring and improving callback plugin performance


dmsimard:
## What is the idea ?

There is a performance overhead between running Ansible with and without ara the [ara callback](https://github.com/ansible-community/ara/blob/master/ara/plugins/callback/ara_default.py) because for every hook, instead of doing nothing, Ansible does this:
![doc/source/_static/graphs/recording-workflow.png](https://github.com/ansible-community/ara/raw/master/doc/source/_static/graphs/recording-workflow.png)

Doing something, even if it's very fast, takes longer than doing nothing... because it adds up when the cost is multiplied by thousands of times across the execution of a larger playbook.
So there will always be an overhead (sorry!) but it should be low enough that it's hopefully worth it to the humans :)

The main issue is that the overhead is exponential due to the multiplication of hosts and tasks.
This means at a small scale it will be low enough to be negligible but a real performance hit when running playbooks involving hundreds of hosts or thousands of tasks.

I'm sure there are improvements we could do to reduce this overhead in the current callback, the API clients or the API server and come up with other implementations with fancy stuff like async, threads, or even dropping the events into a local message bus for lowering the latency to a maximum.

Before we start optimizing stuff, I'd like to measure the overhead even if just for the sake of data, metrics and science.

I think a good start would be:

- raw ansible-playbook
- ansible-playbook with 1.x ara callback enabled
- ansible-playbook with 0.x ara callback enabled (for science and comparison purposes, 0.x doesn't have an API server)

There is a similar issue for the measure API server performance here: https://github.com/ansible-community/ara/issues/170

srgvg:
- remote run with ara disabled: total 63s                                                                           
- remote run with ARA enabled: total 156.55s                                                                        
- local run with ARA enabled: total      118.63s                                                                    
remote being from my laptop at office, local being from a vm in the same network as the targets (cloud DC)        
round trip time from laptop to dc is around 15-20ms

dmsimard:
I'll figure out a synthetic benchmark playbook that we can test with more easily but for the time being, I've tested with two different production-ish playbooks.

The main takeaway that I see for now is that even a 15ms latency between the callback and the API server slows things down quite a bit when compared to a locally-running server -- unless the difference can be attributed to sqlite vs mysql, something that would be easier for me to test with a purpose-built playbook.

The remote server:
- version 1.5.1 (latest stable)
- nginx frontend
[...]

PR #459, opened 2022-11-14
Nonblocking callback

erik-overdahl:
This PR makes it so that the ARA callback module no longer blocks the flow of Ansible playbook execution at all. Now the only blocking happens for the requests sent at the end of the playbook (this is unavoidable). This is accomplished by moving all logic for the callback to a separate thread and pushing received callback messages to the worker thread via a queue.

I ran some very simple benchmarks to confirm that this does in fact improve performance. I would like to have more robust measurements, but right now these are what I had time for.

All measurements were taken with 10 hosts performing 5 tasks across 10 forks. Times are reported in milliseconds and averaged over 5 runs.

First, the test used in [the ARA benchmark article](https://ara.recordsansible.org/blog/2020/11/01/benchmarking-ansible-and-ara-for-fun-and-science/) - each task is just `debug`
|strat|  client|  db|       threads| master| nonblock| diff|
|------| -------| --------| -------| ------| --------| ----- |
|free| | | |                              984| | |
|linear| | | |                            1042| | |
|linear| http|    sqlite|   0|       2628|   2380|     -248|
|free|   http|    sqlite|   0|       3588|   3073|     -515|
|free|   http|    postgres| 4|       4463|   3888|     -575|
|linear| http|    postgres| 0|       4847|   4267|     -580|
|linear| offline| postgres| 4|       4892|   4722|     -170|
|free|   http|    postgres| 0|       7194|   6244|     -950|
|linear| offline| sqlite|   0|       8026|   7426|     -600|
|linear| offline| postgres| 0|       8262|   7487|     -775|
|free|   offline| postgres| 4|       8771|   8039|     -732|
|free|   offline| sqlite|   0|       11835|  10850|    -985|
|free|   offline| postgres| 0|       11939|  10771|    -1168

The second test used the `ansible.builtin.shell` module to sleep for 1 - 3 seconds in order to mimic more realistic conditions
|strat|  client|  db|       threads| master| nonblock| diff|
|------| -------| --------| -------| ------| --------| ----- |
|free| | | |                              15641| | |
|free|   http|    postgres| 4|       15704|  16150|    446| 
|free|   http|    sqlite|   0|       16221|  14842|    -1379 |
|linear| | | |                            18190| | |
|free|   offline| postgres| 4|       18353|  15907|    -2446 |
|free|   http|    postgres| 0|       18454|  15572|    -2882 |
|linear| http|    sqlite|   0|       19229|  18668|    -561|
|free|   offline| sqlite|   0|       19590|  17511|    -2079 |
|linear| http|    postgres| 4|       19919|  19405|    -514|
|free|   offline| postgres| 0|       20170|  16072|    -4098 |
|linear| http|    postgres| 0|       20595|  19226|    -1369 |
|linear| offline| postgres| 4|       20849|  19746|    -1103 |
|linear| offline| postgres| 0|       22578|  20276|    -2302 |
|linear| offline| sqlite|   0|       23304|  19960|    -3344 |

The nonblocking client is faster than the current ARA master branch, and, importantly, makes a big performance difference in the situations where the previous threading-patch cannot help ( using sqlite as the database backend).

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -853,7 +853,7 @@ class CallbackModule(CallbackBase):
         self.queue = Queue()
         # started at the end of set_options()
         self.worker = AraWorker(queue=self.queue)
-        self.worker_thread = threading.Thread(target=self.worker.run)
+        self.worker_thread = threading.Thread(target=self.worker.run, daemon=True)
 
     def set_options(self, task_keys=None, var_options=None, direct=None):
         super().set_options(task_keys=task_keys, var_options=var_options, direct=direct)
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -432,7 +432,10 @@ class AraWorker:
             self._submit_thread("global", self._set_playbook_labels, labels)
 
         # Record all the files involved in the play
-        for path in play._loader._FILE_CACHE.keys():
+        #   make a list of the keys and iterate that for thread safety
+        #   avoiding `RuntimeError: dictionary changed size during iteration`
+        play_files = list(play._loader._FILE_CACHE.keys())
+        for path in play_files:
             # The cache can be pre-populated with files that aren't relevant to the playbook report
             # If there are matches that should be ignored here, don't record them at all
             ignored = False
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -855,14 +855,6 @@ class CallbackModule(CallbackBase):
         self.worker = AraWorker(queue=self.queue)
         self.worker_thread = threading.Thread(target=self.worker.run)
 
-    def __del__(self):
-        """
-        Give the thread as long as possible to complete its work.
-        """
-        self.queue.join()
-        if self.worker_thread.is_alive():
-            self.worker_thread.join()
-
     def set_options(self, task_keys=None, var_options=None, direct=None):
         super().set_options(task_keys=task_keys, var_options=var_options, direct=direct)
 
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -10,7 +10,9 @@ import logging
 import os
 import socket
 import sys
+import threading
 from concurrent.futures import ThreadPoolExecutor
+from queue import Queue
 
 from ansible import __version__ as ANSIBLE_VERSION, constants as C
 from ansible.parsing.ajson import AnsibleJSONEncoder
@@ -253,18 +255,27 @@ ANSIBLE_SETUP_MODULES = frozenset(
 )
 
 
-class CallbackModule(CallbackBase):
+class CallbackSend:
     """
-    Saves data from an Ansible run into a database
+    Same as the one defined in ansible.executor.task_queue_manager;
+    should we just import it?
     """
 
-    CALLBACK_VERSION = 2.0
-    CALLBACK_TYPE = "awesome"
-    CALLBACK_NAME = "ara_default"
+    def __init__(self, method_name, *args, **kwargs):
+        self.method_name = method_name
+        self.args = args
+        self.kwargs = kwargs
 
-    def __init__(self):
-        super().__init__()
-        self.log = logging.getLogger("ara.plugins.callback.default")
+
+class AraQueueDone:
+    pass
+
+
+class AraWorker:
+    def __init__(self, queue: Queue):
+        self.log = logging.getLogger("ara.plugins.callback.worker")
+        self.log.info("NEW ARA")
+        self.queue = queue
         self.localhost_hostname = None
         # These are configured in self.set_options
         self.client = None
@@ -296,60 +307,14 @@ class CallbackModule(CallbackBase):
         self.delegation_cache = {}
         self.warned_about_host_length = []
 
-    def set_options(self, task_keys=None, var_options=None, direct=None):
-        super().set_options(task_keys=task_keys, var_options=var_options, direct=direct)
-
-        self.argument_labels = self.get_option("argument_labels")
-        self.default_labels = self.get_option("default_labels")
-        self.ignored_facts = self.get_option("ignored_facts")
-        self.ignored_arguments = self.get_option("ignored_arguments")
-        self.ignored_files = self.get_option("ignored_files")
-        self.localhost_as_hostname = self.get_option("localhost_as_hostname")
-        self.localhost_as_hostname_format = self.get_option("localhost_as_hostname_format")
-        self.record_controller = self.get_option("record_controller")
-        self.record_user = self.get_option("record_user")
-
-        # The intent for the ignored_files default value is to ignore the ansible local tmpdir but the path
-        # can be changed by the user's configuration so retrieve that and use it instead.
-        # https://github.com/ansible-community/ara/issues/385
-        for pattern in self.ignored_files:
-            if pattern == ".ansible/tmp":
-                tmpdir_config = os.path.dirname(C.config.get_config_value("DEFAULT_LOCAL_TMP"))
-                index = self.ignored_files.index(pattern)
-                self.ignored_files[index] = tmpdir_config
-                break
-
-        client = self.get_option("api_client")
-        endpoint = self.get_option("api_server")
-        timeout = self.get_option("api_timeout")
-        username = self.get_option("api_username")
-        password = self.get_option("api_password")
-        cert = self.get_option("api_cert")
-        key = self.get_option("api_key")
-        ca = self.get_option("api_ca")
-        insecure = self.get_option("api_insecure")
-
-        verify = False if insecure else True
-        if ca:
-            verify = ca
-
-        self.client = client_utils.get_client(
-            client=client,
-            endpoint=endpoint,
-            timeout=timeout,
-            username=username,
-            password=password,
-            cert=cert,
-            key=key,
-            verify=verify,
-        )
-
-        # TODO: Consider un-hardcoding this and plumbing pool_maxsize to requests.adapters.HTTPAdapter.
-        #       In the meantime default to 4 so we don't go above requests.adapters.DEFAULT_POOLSIZE.
-        #       Otherwise we can hit "urllib3.connectionpool: Connection pool is full"
-        self.callback_threads = self.get_option("callback_threads")
-        if self.callback_threads > 4:
-            self.callback_threads = 4
+    def run(self):
+        while True:
+            message = self.queue.get()
+            if isinstance(message, AraQueueDone):
+                self.queue.task_done()
+                return
+            method = getattr(self, message.method_name)
+            method(*message.args, **message.kwargs)
 
     def _submit_thread(self, threadpool, func, *args, **kwargs):
         # Manages whether or not the function should be threaded to keep things DRY
@@ -425,7 +390,8 @@ class CallbackModule(CallbackBase):
         # Record the playbook file
         self._submit_thread("global", self._get_or_create_file, path, content)
 
-        return self.playbook
+        # return self.playbook
+        self.queue.task_done()
 
     def v2_playbook_on_play_start(self, play):
         self.log.debug("v2_playbook_on_play_start")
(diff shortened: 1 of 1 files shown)


dmsimard:
Hey @erik-overdahl and thanks for the PR !

I've looked at the CI failure and I think I know what's going on but it's a bit tricky.

This is the error:
```
TASK [Generate a failure that will be rescued] *********************************
task path: /home/zuul-worker/src/github.com/ansible-community/ara/tests/integration/lookups.yaml:76
fatal: [localhost]: FAILED! => {
    "changed": false,
    "msg": "Message from a failed task"
}

TASK [We can go on, it was a failure for test purposes] ************************
task path: /home/zuul-worker/src/github.com/ansible-community/ara/tests/integration/lookups.yaml:80
ok: [localhost] => {
    "msg": "👍"
}
2022-11-15 17:10:51,558 DEBUG ara.clients.http: HTTP 201: post on /api/v1/tasks
2022-11-15 17:10:51,561 DEBUG ara.plugins.callback.worker: v2_runner_on_start
2022-11-15 17:10:51,561 DEBUG ara.plugins.callback.worker: v2_runner_on_ok

TASK [Find the task with the ignored errors and validate the status] ***********
task path: /home/zuul-worker/src/github.com/ansible-community/ara/tests/integration/lookups.yaml:85
2022-11-15 17:10:51,615 DEBUG ara.clients.http: HTTP 200: get on /api/v1/tasks?name=ignore_errors&playbook=2
fatal: [localhost]: FAILED! => {
    "msg": "The conditional check '_task['results'][0]['name'] == 'Fail something with ignore_errors'' failed. The error was: error while evaluating conditional (_task['results'][0]['name'] == 'Fail something with ignore_errors'): list object has no element 0"
}

PLAY RECAP *********************************************************************
localhost                  : ok=9    changed=1    unreachable=0    failed=1    skipped=1    rescued=1    ignored=1   
```

It's these tasks: https://github.com/ansible-community/ara/blob/dc4344e0707942c5aaf94be6772bdfdab5ae9e2e/tests/integration/lookups.yaml#L71-L91

In other words, we are generating a failure for the sake of testing that we can query and search for it within the playbook.
This is meant to be integration testing coverage for ara itself and I guess it assumes that by the time we query for the failure, it should be there -- but since we're moving stuff around in threads, it could be possible that the failure will get posted later so it can't find it.

I don't have a good suggestion at this time but does my explanation make sense ?

erik-overdahl:
That is my understanding as well. I tried adding a `retry... until... delay`  to that assertion, but when I run the tests locally I am getting all sorts of weird "connection timeout" errors at random and inconsistent points in the test playbooks.

hille721:
Was reading your conversations with much interest. Where I'm a bit afraid of and I feel confirmed after the latest findings is that the reporting is no longer live. As long as the playbook runs to the end everything might be fine. But if the playbook run will be aborted or crashs due to any other reason then there is the possibility that later tasks are already performed before former tasks are fully reported to Ara.

Thus the question of this change is speed vs accuracy. For me speed is important but more important is accuracy. I prefer correct results in Ara then faster playbook runs. (btw that's why I tell everyone in my company to not use the free strategy, because this led to problems in Ara too). 
Are the benchmark results really this good that we want to go for a more complex solution with maybe less accuracy? 
Imo it would be also great to make such a think optional, meaning to support the traditional behavior and making this a feature which you can enable if performance is really important for you.

hille721:
> unless the Ansible program itself crashes

that is exactly what I meant. Maybe a user kills it, or the controller (e.g. a VM) crashs due to any reason, must not be related to the playbook itself. But in such cases I want to have a propper Ara reporting as well.

> The synchronous nature of callbacks in the Ansible architecture...

But wouldn't it better then if directly Ansible supports  kind of async callbacks in general?

erik-overdahl:
If the machine you are running Ansible on crashes, there isn't anything you can do, and current ARA also has the risk of losing results in this case. A task may finish, but if we are still waiting on the POST requests in the callback to finish when the crash happens, then you just lost all results you didn't send yet.

If any exception happens in the Ansible process, the exception is caught at the top level and the process shuts down gracefully - see the [`cli_executor` method](https://github.com/ansible/ansible/blob/devel/lib/ansible/cli/__init__.py#L621). Because the thread runs as a daemon, your outstanding ARA requests will complete before the Ansible process exits. Which, actually, I'm not sure is currently true for ARA? if you cancel the playbook before the request for a task has been sent and you aren't using threads for ARA, then the UserInterrupt error will be thrown and the program will exit and because your requests were happening on that main thread that just got interrupted, you lose that data. So if anything, this PR makes it less likely that you will lose data, not more.

I agree that Ansible SHOULD implement the callback plugins in a non-blocking manner, but they don't, so it is up to the callback plugins themselves to not block.

dmsimard:
> unless the Ansible program itself crashes

> > that is exactly what I meant. Maybe a user kills it, or the controller (e.g. a VM) crashs due to any reason, must not be related to the playbook itself. But in such cases I want to have a propper Ara reporting as well.

> > > If the machine you are running Ansible on crashes, there isn't anything you can do, and current ARA also has the risk of losing results in this case. A task may finish, but if we are still waiting on the POST requests in the callback to finish when the crash happens, then you just lost all results you didn't send yet.

Right.

I might describe ara's attempts to record data as "best effort", to some extent, due to the nature of how the callback interface works. ara's callback gets "woken up" when it receives a hook but if it stops receiving hooks, it can't do anything.
It's why interrupted playbooks and tasks may end up in the "running" status forever because the callback didn't receive the last hook, ``v2_playbook_on_stats``.

There isn't much we can do about that, unfortunately.

> I agree that Ansible SHOULD implement the callback plugins in a non-blocking manner, but they don't, so it is up to the callback plugins themselves to not block.

Over the years, ara has helped with finding [issues in the Ansible callback interface](https://github.com/ansible/ansible/issues?q=is%3Aissue+author%3Admsimard+callback) and while some were fixed, others were closed because fixing them would be too complicated or break stuff.

One of those is about the blocking nature of the callback interface but that was more than 5 years ago now: https://github.com/ansible/ansible/issues/27705

We could revisit that and ask about it upstream but the challenge is that any meaningful change would probably break compatibility and only be available in newer versions of ansible-core since it wouldn't be backported.

Anyway, before we settled on the current threading implementation, every request was synchronous.
This worked reliably well at a small scale but performed poorly at scale because, as you note, ansible would only continue running the playbook once ara's API requests had been sent and returned.

Now, there are still some amount of synchronous requests because, for example, to create a play we need a playbook ID and so we need to make sure the playbook has been created before we try to create the play.

For what it's worth, I went through the exercise of picturing a high level overview of how the sequence currently works from the perspective of the ara callback:
Selection_050

dmsimard:
> Refactoring the callback to be nonblocking addresses more than just performance. The synchronous nature of callbacks in the Ansible architecture means that, in its current form, ARA makes callbacks for things like timing the execution of tasks completely useless. Not only is this bad in itself, but it is also not transparent; people using those callbacks may enable ARA and still be relying on those timings to be accurate representations of the actual task executions times on the hosts, unaware of the slowdown coming from ARA.

I think to call the current durations of tasks ``completely useless`` is a bit of a stretch.
Yes, they include the overhead of ara processing the task because ansible waits for the task to be finished -- but this overhead is typically negligible in the grand scheme of things when tasks can last several minutes long.

There are certain tradeoffs in ara for the sake of simplicity and sometimes it comes at the expense of performance.

Don't get me wrong, performance is important and I understand what you mean and how it would be more accurate to record the task duration as 3m 22s instead 3m 22.5s but that's neither useless or the end of the world.

> Thus the question of this change is speed vs accuracy. For me speed is important but more important is accuracy. I prefer correct results in Ara then faster playbook runs. (btw that's why I tell everyone in my company to not use the free strategy, because this led to problems in Ara too).
> > Avoiding the free strategy because ARA cannot handle it is backwards thinking, imo. If there is a language feature that is perfectly fine by itself but you cannot use it because your logging system cannot handle it, then the problem is with the logging system. ARA should be implemented better. No one should even have to think about including this in their Ansible toolkit and they should not have to change their Ansible playbooks to accommodate ARA.

ara _should_ work reliably with the free strategy. It historically hasn't because I did not personally have a use case for it and so a lot of ara was built around assumptions of running synchronously with the serial strategy.

There are some fixes for the free strategy in 1.6.0 and if there are still issues, we should document and fix them.
I still don't personally have a use case for the free strategy so any help is welcome here.

erik-overdahl:
Yes, upon further reflection, I agree with you that ARA calls being reported in execution time is mostly a non-issue. I was thinking of it as reporting the execution time of a task on a host, but that is plainly not the case when the number of hosts is greater than `ANSIBLE_FORKS`.

I personally think that the performance gains from this patch are worth it, but I understand if you do not want to take on the added complexity. It isn't much more complexity, but it isn't none either. I can try to update that integration test to wait until the record is available, if you like (although I'm not sure of the best way to do this - you can't really retry until the record appears, because the test would never fail if the record wasn't actually there, so then your option is, what, a timeout? but that seems flaky, especially in a test, which you would hope is 100% deterministic).

Perhaps the most useful datapoint on ARA performance is [this comment from the issue related to the benchmark blog post](https://github.com/ansible-community/ara/issues/171#issuecomment-709928641) which claims that the 30% gains they saw from using Ansible with Mitogen were wiped out by ARA.

I would like to have a wider range of benchmarks. The numbers in the original ARA blogpost make the situation look a lot worse than it actually is. But I think it would be difficult to develop something useful; there are a lot of potential scenarios in a realistic workload.

Re: the free strategy - One of the things I found when I sprinkled logging statements throughout the callback code was that the ARA threading strategy ends up killing and creating a lot of ThreadPoolExecutors as the `free` strategy calls `v2_playbook_on_task_start` when it schedules a task on a host. In the ARA callback module, this triggers the `_end_task` method, as ARA assumes the linear strategy in which all runners for a task complete before the next task begins. Oddly, though... I don't know if this is actually an issue?? Look at the benchmarks in my first post; the free strategy with multiple threads still seems to perform pretty well.

dmsimard:
Hi @erik-overdahl,

I just wanted to say I haven't forgotten about your PR and that I'll get back to it when the time is right.
I believe the callback is due for a refactor but I don't know what that would look like in practice yet.

The callback has grown organically over the years: it has had features and fixes patched on along some reverse engineering of internal ansible-core's "API"s. It could be a good idea to take a step back, look at what it does and what we've learned over the years to come up with something a bit cleaner.

ara didn't always have an API and the callback used to make database queries directly with flask so a lot has changed since then :p

Performance, threading, queueing or asynchronicity can be part of an eventual refactor but I mean to say that a lot of what the callback does could benefit from being put in a library -- things like creating a playbook, a task, a host or a result, for example.

I want to have a better idea of what this looks like before going down the road of optimizing performance such that we don't step on each other's feet.

I appreciate your patience in the meantime !

phemmer:
If the risk of losing updates on crash is an issue, I think we can address those easily enough.

As @erik-overdahl mentioned we should be able to delay the shutdown until the queue drains. A typical approach I see in other applications (including my own) with similar concerns is that SIGINT/SIGTERM will initiate a graceful shutdown (wait for things to terminate cleanly, and in this case for the queue to drain), and a second SIGINT/SIGTERM causes an immediate exit.

We can also put in a max queue size. For example if we make the max size 0, then the request has to be passed synchronously to the thread, but once the thread has it ansible can continue and the thread would make the API call in the background. Any further attempts to push something onto the queue will block until the API call completes.
This can be customizable, allowing the user to specify how big that max size should be based on how much risk of lost data they're willing to accept. And a max size of -1 would disable async entirely (though might make it more intuitive to shift the values by +1, so 0 would disable async).

I also just want to add my own real world experience. The benchmarks in the original post were overly optimistic for my experience. My playbooks have lots of very short tasks that run across hundreds of hosts. I saw playbooks taking over 3x longer when ARA was enabled. A playbook that took about 7 minutes was taking almost 30 minutes. Though this was a while ago, and I cant remember details. I might have been using django auth, and might have been using sqlite, of which I now use neither, and am using the async wrapper I posted in https://github.com/ansible-community/ara/issues/171#issuecomment-1302863591.

berendt:
I have tested the Async Wrapper. It does not work for us because it changes the timestamps in ARA (or in other words, these do not correspond to the actual time of execution). But I think the idea of the async wrapper (thanks for sharing @phemmer) is a good one.

dmsimard:
I have not given up on this PR (and more broadly improving callback performance) but the project is Migrating to Codeberg.

It has been migrated here: https://codeberg.org/ansible-community/ara/pulls/459

I am thus closing this pull request since merges will occur on Codeberg and will be mirrored back to GitHub.

PR #460, opened 2022-11-20
tests: make it easier to run with_container_images.yml locally

dmsimard:
- Make the default ara_api_source compatible whether it's running locally or from Zuul
- Only gather the facts we need
- Don't try to elevate privileges if we don't need to
- Don't assume running with selinux (this can be tested elsewhere)


dmsimard:
I've squashed this into https://github.com/ansible-community/ara/pull/491 so it would be easier for me to test locally.

PR #470, opened 2022-12-12
fix callback plugin creates broken folder in the current working directory #469

### LINKED ISSUE #469: callback plugin creates broken folder in the current working directory
## What is the issue ?
When running any playbook with ARA callback plugin enabled a malformed directory is created in the same folder as ansible.cfg.  This only happens when ARA (1.6.0) is enabled.

```console
$ ls -1
ansible.cfg
'{{ ANSIBLE_HOME ~ "'
bootstrap.sh
collections
container.def
inventory
playbooks
README.md
requirements.txt
roles
web_services

$ tree \{\{\ ANSIBLE_HOME\ ~\ \"/
{{ ANSIBLE_HOME ~ "/
└── tmp" }}
```

If I delete the directory, disable ARA, and run playbook again it isn't recreated.  Then when I enable ARA the same playbook will create that strange directory.

playbook
```yaml
---

- name: "Test Ansible Connection"
  hosts: "{{ cplay_hosts | default('all') }}"
  become: false
  strategy: linear
  gather_facts: no
  tasks:
    - name: connection test
      ping:
  post_tasks:
    - name: stop ssh client persistant connection
      meta: reset_connection
```

## What should be happening ?
Additional directories should not be created in the same folder as ansible.cfg as this can make git mad.

## Software Versions
Ansible = 7.1.0 (Core = 2.14)
OS = Fedora 37
Python = 3.11
ARA = 1.6.0
Install Source = pip

hille721:
fixes #469

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -314,7 +314,7 @@ class CallbackModule(CallbackBase):
         # https://github.com/ansible-community/ara/issues/385
         for pattern in self.ignored_files:
             if pattern == ".ansible/tmp":
-                tmpdir_config = os.path.dirname(C.config.get_config_value("DEFAULT_LOCAL_TMP"))
+                tmpdir_config = os.path.dirname(C.DEFAULT_LOCAL_TMP)
                 index = self.ignored_files.index(pattern)
                 self.ignored_files[index] = tmpdir_config
                 break


dmsimard:
@hille721 I've asked #ansible-devel about it to confirm the behavior and waiting to hear back.

Did you test different versions of ansible ?
The CI jobs are green but I'd like to double check that the behavior is consistent across versions of ansible.

Over the years it hasn't been uncommon to manage different versions of ansible differently...

hille721:
I tested only with 2.13 and 2.14. 
Also unsure about the "offical" way to use these constants. But I think it's at least better than before.

dmsimard:
The jobs are green and there's an integration test added just for that: https://github.com/ansible-community/ara/blob/master/tests/integration/ignored_files.yaml

I still double checked that it really worked and it does, at least as far back as 2.9:
Screenshot from 2022-12-12 22-25-56

Thanks for the fix @hille721 !

I will prepare a release of 1.6.1 so we can ship it.

PR #473, opened 2022-12-21
ui: truncate long controller hostnames

dmsimard:
Fields that can be very long are generally truncated in tables with the full version available on mouse-over.

This is a compromise to avoid a single column taking up most of the width available in the table.

controller hostnames can be up to 255 characters so truncate them like we do elsewhere.


dmsimard:
Before:
Screenshot from 2022-12-21 17-45-08

After:
Screenshot from 2022-12-21 18-32-25

PR #475, opened 2023-01-26
Missing end of quote

RicardoM17:


dmsimard:
Indeed. Thanks for the fix @RicardoM17 !

RicardoM17:
No problem @dmsimard , glad to make contributions, even if small :D

P.S. I'm starting to use Ansible and ARA, and ARA is awesome!

dmsimard:
I will take small (but valid!) contributions like this any time.

Thanks for the nice words and I'm glad you like it!

PR #496, opened 2023-06-29
Allow manually defining user & hostname in callback plugin

### LINKED ISSUE #493: Allow specifying username to record
## What is the idea ?

Currently ARA pulls the username from the OS user the process is running as. In my setup, ansible is run via CI jobs, and also from a container, so the user is always recorded as "root", which is not useful. The CI (GitLab) provides a variable which indicates the user who triggered the job, and I would like to be able to use this variable to populate the recorded user.

It seems like this change should be simple enough by adding a new option and using the value if set, otherwise falling back to the current behavior. I'd be willing to submit a PR if you agree, and have an idea on what the option should be named.

&nbsp;

On a side note, host name also suffers from the same issue, as it gets the name of the container. However this is not important to me, but it may be important to others.

phemmer:
In some environments (e.g. containers), the user and hostname detected from the OS might not be relevant or useful. This change allows the user to pass in customized values for these attributes which override the automatic detection.

Note: I didn't see any tests for things like the the the controller name or user name, or any other playbook level attributes, so I didn't add any here.

Closes #493

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -227,6 +227,16 @@ options:
     ini:
       - section: ara
         key: record_controller
+  record_controller_name:
+    description:
+      - The name to use when recording the controller hostname.
+      - Defaults to the system hostname.
+    type: string
+    env:
+      - name: ARA_RECORD_CONTROLLER_NAME
+    ini:
+      - section: ara
+        key: record_controller_name
   record_user:
     description:
       - Whether ara should record the user that ran a playbook
@@ -238,6 +248,16 @@ options:
     ini:
       - section: ara
         key: record_user
+  record_user_name:
+    description:
+      - The name to use when recording the user that ran a playbook.
+      - Defaults to the OS user which ran the playbook.
+    type: string
+    env:
+      - name: ARA_RECORD_USER_NAME
+    ini:
+      - section: ara
+        key: record_user_name
 """
 
 # Task modules for which ara should save host facts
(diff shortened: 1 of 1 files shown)


ara/plugins/callback/ara_default.py
```diff
@@ -848,6 +873,9 @@ def _get_user(self):
         if not self.record_user:
             return user
 
+        if self.record_user_name:
+            return self.record_user_name
```
dmsimard, line 877: I wonder for both the user and the controller, should we move this before the first if ?

This way setting ``ARA_RECORD_USER_NAME`` would be sufficient without requiring ``ARA_RECORD_USER`` to be true.

dmsimard:
Thanks for the PR!

> In some environments (e.g. containers), the user and hostname detected from the OS might not be relevant or useful. This change allows the user to pass in customized values for these attributes which override the automatic detection.

Looks good to me, though I had a question in-line.
 
> Note: I didn't see any tests for things like the the the controller name or user name, or any other playbook level attributes, so I didn't add any here.

There's unit tests for playbook properties but it's more about the API than the callback plugin since a lot of it is mocked: https://github.com/ansible-community/ara/blob/master/ara/api/tests/tests_playbook.py

For testing playbook properties as recorded by the callback we have integration tests written as Ansible playbooks like here: https://github.com/ansible-community/ara/blob/95176a9ef1fad756ef4a898d068497fdd7f086f5/tests/integration/lookups.yaml#L13-L41

That particular playbook is one of the test playbooks launched when running ``tox -e ansible-integration``.

Adding coverage is always nice but not a hard requirement since we can take care of it in a follow up PR before release.

ara/plugins/callback/ara_default.py
dmsimard: On second thought, these settings already default to true so it would only matter if someone explicitly set them to false.

It's probably fine like this, what do you think ?

phemmer: Yes, that was my thought. It's enabled by default, and if someone wants to disable it, then disabling would apply to the custom specification as well.

dmsimard:
Hi @phemmer and sorry it's taken so long.

There is nothing wrong with the code and I've tested it locally successfully:
Screenshot from 2023-09-03 15-15-21

I will merge this as-is and follow up with a patch for integration testing it.

dmsimard:
This will be released as part of ara 1.7.0 sometime in the near future.

Issue #506, opened 2023-09-03
Write integration tests for new ARA_RECORD_ settings


dmsimard:
## What is the idea ?

The callback has grown three new settings thanks to the contributions in https://github.com/ansible-community/ara/pull/458 and https://github.com/ansible-community/ara/pull/496:
- ARA_RECORD_USER_NAME (to manually override the recorded user for a playbook)
- ARA_RECORD_CONTROLLER_NAME (to manually override the recorded controller for a playbook)
- ARA_RECORD_TASK_CONTENT (to prevent ara from recording task content)

Before releasing 1.7.0, there needs to be a minimum of testing coverage for them.

It might end up looking similar to what we do when testing controller hostnames: https://github.com/ansible-community/ara/blob/d736213f5569df888e25cbb77161158b94e18fcd/tests/basic.yaml#L191-L194

And then, in something like the [lookups test playbook](https://github.com/ansible-community/ara/blob/master/tests/integration/lookups.yaml), we'd add tests to validate that we have indeed the expected name, hostname and that we don't have the task content.

PR #507, opened 2023-09-04
callback: add tests for ARA_RECORD_* settings

dmsimard:
There are three new settings which needed some amount of testing coverage:
- ARA_RECORD_USER_NAME (to manually override the recorded user for a playbook)
- ARA_RECORD_CONTROLLER_NAME (to manually override the recorded controller for a playbook)
- ARA_RECORD_TASK_CONTENT (to prevent ara from recording task content)

This improves the coverage while fixing a gap that we had about controller and user name recording.

Fixes: https://github.com/ansible-community/ara/issues/506

### DIFF
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -139,11 +139,3 @@
       - name: github.com/ansible/ansible
         override-checkout: stable-2.14
 
-- job:
-    name: ara-basic-ansible-2.9
-    parent: ara-ansible-integration-base
-    description: |
-      Runs basic integration tests through the equivalent of "tox -e ansible-integration" with Ansible 2.9.
-    required-projects:
-      - name: github.com/ansible/ansible
-        override-checkout: stable-2.9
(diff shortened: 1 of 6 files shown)


dmsimard:
The issue with 2.9 is because this (perhaps [elaborate](https://github.com/ansible-community/ara/pull/507/files#diff-2f2ceae9c31b4dab697c7c3bc8ff197205854bdc647f75085131bd769b140028R31-R40)) loop behaves differently which is not particularly surprising given the number of differences since 2.9 was released.

In other words, the feature works fine and is tested successfully against 2.14 and 2.15 but we've kept 2.9 in the testing CI since it was a good "snapshot" before the ansible-core era with ansible collections.

I think it is reasonable to stop testing against 2.9 considering it's been [EOL for over a year now](https://groups.google.com/g/ansible-announce/c/kegIH5_okmg/).

dmsimard:
It works:
Screenshot from 2023-09-09 01-45-25

Issue #499, opened 2023-07-29
UI: playbooks with many labels (or long labels) messes with the table layout


dmsimard:
## What is the issue ?

Tables are easier to read when everything is on a single line. Yet, when there are many labels or even just a few long labels, it throws off the table layout like this:

![256712010-cc735df5-4791-4446-b68a-317a2c637906](https://github.com/ansible-community/ara/assets/1291204/82863e3c-190e-418d-bc47-9ea9bac5d19e)

It's not great.

## What should be happening ?

Displaying the labels are meant to provide an easy way to filter playbooks by labels by clicking on a particular label.
Displaying all of them in their entirety messes with the table layout, though.

We should find some form of implementation that improves on the current setup.

PR #509, opened 2023-09-09
ui: move playbook labels onto their own row

dmsimard:
This reduces the waste of whitespace from the labels wrapping.

Related: https://github.com/ansible-community/ara/issues/499

### DIFF
--- a/ara/ui/static/css/ara.css
+++ b/ara/ui/static/css/ara.css
@@ -35,8 +35,8 @@
 
 /* Size playbook labels and add padding/margin */
 .ara-label {
-    font-size: 90% !important;
-    padding: 0.10rem 0.25rem !important;
+    font-size: 95% !important;
+    padding: 0.05rem 0.25rem !important;
     margin-bottom: 3px;
 }
 
@@ -76,6 +76,7 @@
 .report-column {
     width: 60px;
     text-align: center;
+    vertical-align: middle;
 }
 
 .short-status-column {
--- a/ara/ui/templates/partials/tables/playbooks.html
+++ b/ara/ui/templates/partials/tables/playbooks.html
@@ -4,7 +4,7 @@
 {% load static_url %}
 
 <div class="table-responsive">
-  <table class="table table-sm table-hover" id="playbooks">
+  <table class="table table-sm table-hover table-striped" id="playbooks">
     <thead>
       <tr style="height:50px;">
         <th class="report-column" title="Link to the playbook report">Report</th>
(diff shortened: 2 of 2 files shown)


dmsimard:
I'm not super confident it's the prettiest implementation but it looks better ¯\_(ツ)_/¯

## Before

Screenshot from 2023-09-09 02-53-54

Screenshot from 2023-09-09 02-53-57

## After

Screenshot from 2023-09-09 03-34-25

Screenshot from 2023-09-09 03-34-21

dmsimard:
If we go ahead with something like this, it would be worth revisiting the hosts and tasks tables to do something similar with host facts and task tags:

Screenshot from 2023-09-09 03-37-53

Screenshot from 2023-09-09 03-37-48

dmsimard:
@ianw, @hille721, what do you think ? do you have an opinion ?

hille721:
from only checking the pictures it looks a bit overloaded. But generally I thing it's a great idea. What could be help is to better sperate the rows, eg. by a clearer borderline or a bit of whitespace between the rows.

dmsimard:
> from only checking the pictures it looks a bit overloaded. But generally I thing it's a great idea. What could be help is to better sperate the rows, eg. by a clearer borderline or a bit of whitespace between the rows.

... I agree. I think if we could have everything on one line it would be ideal and it's part of why I've been trying to resist a change like this. I thought it was worth trying out even if just to visualize what it looked like in practice.

There's too much information for the amount of width we have available which leads us to truncating strings and resorting to these "hacks".. UI/UX is hard and I might put this off for now unless we figure out what would look good.

dmsimard:
Yeah, I'll leave this open but remove it from the 1.7.0 milestone.

This isn't a blocker for the release and we can revisit it later.

ianw:
The tags contain a lot of repeated information and probably aren't that helpful for "first level" debugging ... i.e. my playbook usually works but something broke; they seem more useful for the less common but still relevant "something is subtly wrong with how this this working".   

I know that it's easy to say and not so easy to write up in javascript; it feels like maybe if there was a tags label with a count that you mouse-over and it unrolls the tags, click to keep them around, and a toggle "show all tags" it would be the best of all worlds.

One thing about the sample https://5636c233c25b7ab53fbc-f7539d74592ba1909ca4d781e774dc92.ssl.cf5.rackcdn.com/509/066ab7e3bea8c31580772604d54fae712c08db60/check/ara-basic-ansible-core-2.15/0fb611f/server/static/index.html is that the labels are a button, so highlights and gives a pointer when moving over, suggesting it is clickable, which it really isn't?

dmsimard:
> The tags contain a lot of repeated information and probably aren't that helpful for "first level" debugging ... i.e. my playbook usually works but something broke; they seem more useful for the less common but still relevant "something is subtly wrong with how this this working".

I agree.

> I know that it's easy to say and not so easy to write up in javascript; it feels like maybe if there was a tags label with a count that you mouse-over and it unrolls the tags, click to keep them around, and a toggle "show all tags" it would be the best of all worlds.

My skills and interest in frontend are limited, pulls requests will be considered :D

> One thing about the sample https://5636c233c25b7ab53fbc-f7539d74592ba1909ca4d781e774dc92.ssl.cf5.rackcdn.com/509/066ab7e3bea8c31580772604d54fae712c08db60/check/ara-basic-ansible-core-2.15/0fb611f/server/static/index.html is that the labels are a button, so highlights and gives a pointer when moving over, suggesting it is clickable, which it really isn't?

Oh, that's because in the non-static version they are clickable and allows to search across every playbook for a label -- i.e: https://dev.demo.recordsansible.org/?label=tags:keystone

If we really wanted to fix that, it'd be inside the various if's like this one, I suppose we could have a different class or something:
https://github.com/ansible-community/ara/blob/9cadcbb2fcc66b10969670fe7f98fe8511a9a2f6/ara/ui/templates/partials/tables/playbooks.html#L144-L150

hille721:
> The tags contain a lot of repeated information and probably aren't that helpful for \"first level\" debugging 

Are we talking about tags or labels?
If we are talking about labels, then I disagree, as a user you can define what should be in the labels. E.g. I only have one to two lables, which are mostly grouping my playbooks. Thus the label directly tells me to wich category the playbook belongs and that it is a clickable button to filter for it, is really great.

Issue #514, opened 2023-09-11
1.7.0rc3 - ModuleNotFoundError: No module named 'ara'


hille721:
I just wanted to rest the new ara version 1.7.3 but ara is not starting with `ModuleNotFoundError: No module named 'ara'` error.

This is the full output when starting ara:

```
[ara] No setting found for SECRET_KEY. Generating a random key...
[ara] Writing default settings to /opt/ara/settings.yaml
[ara] Using settings file: /opt/ara/settings.yaml
Operations to perform:
  Apply all migrations: admin, api, auth, contenttypes, db, sessions
Running migrations:
  Applying contenttypes.0001_initial... OK
  Applying auth.0001_initial... OK
  Applying admin.0001_initial... OK
  Applying admin.0002_logentry_remove_auto_add... OK
  Applying admin.0003_logentry_add_action_flag_choices... OK
  Applying api.0001_initial... OK
  Applying api.0002_remove_host_alias... OK
  Applying api.0003_add_missing_result_properties... OK
  Applying api.0004_duration_in_database... OK
  Applying api.0005_unique_label_names... OK
  Applying api.0006_remove_result_statuses... OK
  Applying api.0007_add_expired_status... OK
  Applying api.0008_playbook_controller... OK
  Applying api.0009_latesthost... OK
  Applying api.0010_result_delegated_to... OK
  Applying api.0011_play_label_name_max_length... OK
  Applying api.0012_playbook_user... OK
  Applying api.0013_task_status... OK
  Applying api.0014_ara_versions... OK
  Applying api.0015_task_uuid... OK
  Applying api.0016_revert_play_label_name_length... OK
  Applying api.0017_optional_playbook_controller... OK
  Applying contenttypes.0002_remove_content_type_name... OK
  Applying auth.0002_alter_permission_name_max_length... OK
  Applying auth.0003_alter_user_email_max_length... OK
  Applying auth.0004_alter_user_username_opts... OK
  Applying auth.0005_alter_user_last_login_null... OK
  Applying auth.0006_require_contenttypes_0002... OK
  Applying auth.0007_alter_validators_add_error_messages... OK
  Applying auth.0008_alter_user_username_max_length... OK
  Applying auth.0009_alter_user_last_name_max_length... OK
  Applying auth.0010_alter_group_name_max_length... OK
  Applying auth.0011_update_proxy_permissions... OK
  Applying auth.0012_alter_user_first_name_max_length... OK
  Applying db.0001_initial... OK
[...]

Issue #406, opened 2022-08-05
container-images CI job suddenly failing with Connection refused when testing image builds


dmsimard:
## What is the issue ?

The ara-container-images job has suddenly started consistently failing: https://ansible.softwarefactory-project.io/zuul/builds?job_name=ara-container-images&skip=0

![Screenshot from 2022-08-05 16-44-47](https://user-images.githubusercontent.com/1291204/183160342-caaa0d4e-4a5f-4970-a2c1-580d8041bf0d.png)

Some logs:

```
2022-08-05 16:38:11.438194 | TASK [Build localhost/ara-api:fedora35-source-latest with fedora-source.sh]
2022-08-05 16:38:11.792480 | fedora-36 | +++ dirname /home/zuul-worker/src/github.com/ansible-community/ara/contrib/container-images/fedora-source.sh
2022-08-05 16:38:11.794549 | fedora-36 | ++ cd /home/zuul-worker/src/github.com/ansible-community/ara/contrib/container-images
# ...
2022-08-05 16:39:16.004294 | fedora-36 | + buildah config --env ARA_BASE_DIR=/opt/ara fedora-working-container
2022-08-05 16:39:16.069974 | fedora-36 | + buildah config --cmd 'bash -c '\''/usr/local/bin/ara-manage migrate && /usr/bin/gunicorn-3 --workers=4 --access-logfile - --bind 0.0.0.0:8000 ara.server.wsgi'\''' fedora-working-container
2022-08-05 16:39:16.136068 | fedora-36 | + buildah config --port 8000 fedora-working-container
2022-08-05 16:39:16.202359 | fedora-36 | + buildah commit fedora-working-container localhost/ara-api:fedora35-source-latest
2022-08-05 16:39:17.267884 | fedora-36 | Getting image source signatures
2022-08-05 16:39:17.268497 | fedora-36 | Copying blob sha256:bed35f21a436df75dbcd6e18a7552084e9a4dd15ee0fb3dfc0bc82d1b73a7586
2022-08-05 16:39:17.268782 | fedora-36 | Copying blob sha256:e55023378e18fdd3329296df3655533630fa9b7dab5228b0549c826ac6126c6e
2022-08-05 16:39:20.893704 | fedora-36 | Copying config sha256:f5550890a0801ef52bbde176613e186eb056208c80492a561f2677918039bc28
2022-08-05 16:39:20.915395 | fedora-36 | Writing manifest to image destination
[...]

PR #515, opened 2023-09-11
python3 -m gunicorn instead of /usr/local/bin/gunicorn as container entry command

hille721:
To avoid import problems such as in https://github.com/ansible-community/ara/issues/406#issue-1330390967 or #514 it is more safe to call `gunicorn` via `python3 -m` same as it is done with `pip`


dmsimard:
Looks good to me, thanks for also fixing those outdated lines :)

Issue #524, opened 2023-10-24
ruamel.yaml 0.18 breaks ara


noonedeadpunk:
## What is the issue ?

With release of ruamel.yaml 0.18 ara fails the playbook execution with the following error in output:

```
2023-10-24 07:47:54.494053 | ubuntu-jammy | /opt/ansible-runtime/lib/python3.10/site-packages/ara/server/settings.py:306: DeprecationWarning:
2023-10-24 07:47:54.635779 | ubuntu-jammy | dump has been removed, use
2023-10-24 07:47:54.635841 | ubuntu-jammy |
2023-10-24 07:47:54.635849 | ubuntu-jammy |   yaml=YAML(typ='unsafe', pure=True)
2023-10-24 07:47:54.635857 | ubuntu-jammy |   yaml.dump(...)
2023-10-24 07:47:54.635863 | ubuntu-jammy |
2023-10-24 07:47:54.635868 | ubuntu-jammy | instead of
2023-10-24 07:47:54.635874 | ubuntu-jammy |
2023-10-24 07:47:54.635899 | ubuntu-jammy |   yaml.dump({"default": SETTINGS}, settings_file, default_flow_style=False)
2023-10-24 07:47:54.635911 | ubuntu-jammy | [ara] No setting found for SECRET_KEY. Generating a random key...
2023-10-24 07:47:54.635918 | ubuntu-jammy | [ara] Creating data & configuration directory: /root/.ara/server
2023-10-24 07:47:54.635924 | ubuntu-jammy | [ara] Writing default settings to /root/.ara/server/settings.yaml
2023-10-24 07:47:54.635939 | ubuntu-jammy |
```

Bringing version of the package <0.18 solves the issue.

dmsimard:
Hi and thanks for reporting the issue @noonedeadpunk o/

As you pointed out, ruamel.yaml 0.18 was released recently and introduced the change:
![Screenshot from 2023-10-24 16-13-50](https://github.com/ansible-community/ara/assets/1291204/3e709b73-3576-48dc-9225-40974090fa1e)

I checked https://review.opendev.org/c/openstack/openstack-ansible/+/899196 and it's good to know that pinning ruamel.yaml is a valid workaround for now.

We have this note on our usage of ruamel.yaml: https://github.com/ansible-community/ara/blob/dde76744596b031f8a52297a71fee59ba9aeb605/ara/server/settings.py#L12-L17

We'll need to fix this part that writes out the default configuration file:
[...]

PR #538, opened 2024-02-06
settings: fix deprecated use of .dump in ruamel.yaml

dmsimard:
ruamel.yaml>=0.18 deprecated the use of .dump which we historically used that maintained compatibility between pyyaml and ruamel.yaml.

We can eventually remove this pin when we change our usage of dump.

https://github.com/ansible-community/ara/issues/524

### DIFF
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -2,21 +2,21 @@
 - job:
     name: ara-tox-py3
     parent: tox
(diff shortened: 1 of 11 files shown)


dmsimard:
The container images job is failing because there needs to be a release of ara 1.7.1 to pypi first.
I've temporarily disabled the image publishing for now and will re-enable it later.

PR #540, opened 2024-03-05
ui: fix displaying result diffs with "prepared" key

mk-fg:
This fixes diffs in task results that only have "prepared" key in them from looking like this:

scr-20240305002825

To being properly formatted and colorized, like diffs with "before" and "after" keys in them look.

Ansible produces diffs with "prepared" key instead of before/after in plugins like git, apt, cli_config already.

But also some network devices (e.g. Arista EOS, Cisco IOS, etc) can only produce diffs like this between normalized candidate and running configurations, where it is tricky or impossible to do local diff, without on-device knowledge about configuration format normalization.
(I've bumped into this issue with arista.eos ansible collection in particular)

Thanks.

### DIFF
--- a/ara/ui/templatetags/diff_result.py
+++ b/ara/ui/templatetags/diff_result.py
@@ -9,13 +9,16 @@ from django import template
 register = template.Library()
 
 
-def render_diff(before="", after="", before_header="before", after_header="after"):
+def render_diff(before="", after="", prepared="", before_header="before", after_header="after"):
     """
     Renders a diff provided by Ansible task results
     """
     # fmt: off
-    # Some modules, such as file, might provide a diff in a dict format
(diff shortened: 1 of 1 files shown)


dmsimard:
Hi @mk-fg and thanks for the PR. This looks like a good patch.

I have been away for a bit and catching up, appreciate the patience.

I want to ask around a bit to find out if there are any other such cases that we should know about or if there is some sort of unwritten convention around ``prepared``. We can probably merge this when I hear back.

dmsimard:
Hi, it's been a while but I did check and found out that ``prepared`` is a known convention.

I was curious whether other known variables would need a similar treatment but it doesn't seem to be the case.

We can merge this, thanks for the PR @mk-fg :)

PR #542, opened 2024-03-27
Support AWS Fargate env key/value schema

osulli:
...  that only allows strings by supporting the ability to pass a comma-delimited string e.g. "1,127.0.0.1,localhost"

### DIFF
--- a/ara/server/settings.py
+++ b/ara/server/settings.py
@@ -69,7 +69,10 @@ if "root" in LOGGING:
     del LOGGING["root"]
 
 # Django built-in server and npm development server
-ALLOWED_HOSTS = settings.get("ALLOWED_HOSTS", ["::1", "127.0.0.1", "localhost"])
+if isinstance(settings.get("ALLOWED_HOSTS"), str):
+    ALLOWED_HOSTS = settings.get("ALLOWED_HOSTS", "").split(',')
+else:
+    ALLOWED_HOSTS = settings.get("ALLOWED_HOSTS", ["::1", "127.0.0.1", "localhost"])
 CORS_ORIGIN_WHITELIST = settings.get("CORS_ORIGIN_WHITELIST", ["http://127.0.0.1:8000", "http://localhost:3000"])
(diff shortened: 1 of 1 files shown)


dmsimard:
Hi @osulli, I am not familiar with Fargate but out of curiosity, does it work if you supply the variable as shown in the docs ?

Per https://ara.readthedocs.io/en/latest/api-configuration.html#ara-allowed-hosts:
``export ARA_ALLOWED_HOSTS="['api.ara.example.org', 'web.ara.example.org']"``

osulli:
> Hi @osulli, I am not familiar with Fargate but out of curiosity, does it work if you supply the variable as shown in the docs ?
> 
> Per https://ara.readthedocs.io/en/latest/api-configuration.html#ara-allowed-hosts:
> ``export ARA_ALLOWED_HOSTS="['api.ara.example.org', 'web.ara.example.org']"``

Unfortunately, AWS Fargate (a managed K8s service) does not support arrays but only strings for env vars.
Instead of building this fix and pushing it to a repository I decided to just create a server and ssh into it and run the docker commands 😂

dmsimard:
> > Hi @osulli, I am not familiar with Fargate but out of curiosity, does it work if you supply the variable as shown in the docs ?
> > Per https://ara.readthedocs.io/en/latest/api-configuration.html#ara-allowed-hosts:
> > `export ARA_ALLOWED_HOSTS="['api.ara.example.org', 'web.ara.example.org']"`
> 
> Unfortunately, AWS Fargate (a managed K8s service) does not support arrays but only strings for env vars. Instead of building this fix and pushing it to a repository I decided to just create a server and ssh into it and run the docker commands 😂

The value of ARA_ALLOWED_HOSTS in the example above is a string, though.
It gets parsed as a python list but the value itself is a string: ``['api.ara.example.org', 'web.ara.example.org']`` so I was wondering if that worked or not.

flowerysong:
*All* environment variables are strings. The question is, does Fargate prevent you from setting certain strings? There's nothing in the documentation that suggests that it does, so you should be able to set it to the correct string instead of modifying ara's config parsing.

Issue #543, opened 2024-04-29
"Duration" column on the playbook view page doesn't sort


evgeni:
## What is the issue ?

1. go to https://demo.recordsansible.org/playbooks/3991.html
2. click on the duration column to sort by duration
3. expect the table to be sorted by duration now, while it is not

## What should be happening ?

It should sort ;-)

Refs: https://chaos.social/@zhenech/112355188740814889

dmsimard:
I looked and the button is there as intended: https://github.com/ansible-community/ara/blob/c750fcd5a7e62f0b9087c05f219c9f8ba6d7dbf9/ara/ui/templates/partials/tables/playbook_results.html#L31

When comparing it to the playbook list table that works, I notice that the ``<form>`` is there: https://github.com/ansible-community/ara/blob/c750fcd5a7e62f0b9087c05f219c9f8ba6d7dbf9/ara/ui/templates/partials/tables/playbooks.html#L16-L18

While it's missing for the task table. That's probably the issue.

dmsimard:
The sort by date field has the same issue as well.

dmsimard:
@evgeni https://github.com/ansible-community/ara/pull/544 works for me and should fix it.

Thanks for the issue :)

evgeni:
FWIW, it works for me too now :)

PR #544, opened 2024-04-29
ui: Fix sorting playbook results by date and duration

dmsimard:
It used to be that sort_by_date and sort_by_duration accepted a "form_url" argument before 1.7.0 but it is no longer the case.

Like the playbook list, use an in-line form element instead.

Fixes: https://github.com/ansible-community/ara/issues/543


dmsimard:
Tested locally and confirmed that it works on the live demo as well.

PR #548, opened 2024-05-25
Docs: Add introduction, say what ARA does

kpinc:
Copied from the FAQ, "Why ARA and not <X>"

(It's nice to have the docs begin by explaining what the software does.  :)


dmsimard:
I like to think the README does a good job for this but I guess the README isn't officially reproduced in the docs and this doesn't hurt so we can merge it. Thanks :)

PR #550, opened 2024-06-12
setup.cfg: make ruamel.yaml dependency explicit

dmsimard:
It should generally be picked up by the dynaconf yaml extras but it was found to not always be the case (i.e: Fedora koji) so make it explicit since there is no harm in doing so.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -74,6 +74,9 @@ server=
     # dynaconf 3.1.3 had a regression https://github.com/ansible-community/ara/issues/212
     # dynaconf dropped support for py35 at version 3.0.0: https://github.com/ansible-community/ara/issues/370
     dynaconf[yaml]>=3.0.0,!=3.1.3,<4.0.0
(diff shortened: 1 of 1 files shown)


dmsimard:
Before ruamel was made explicit: https://koji.fedoraproject.org/koji/taskinfo?taskID=118918621
```
  File "/builddir/build/BUILD/ara-1.7.2.dev5-build/ara-1.7.2.dev5/ara/server/settings.py", line 16, in <module>
    from ruamel.yaml import YAML
ModuleNotFoundError: No module named 'ruamel'
RPM build errors:
error: Bad exit status from /var/tmp/rpm-tmp.5xX5nS (%check)
    Bad exit status from /var/tmp/rpm-tmp.5xX5nS (%check)
Child return code was: 1
```

After making it explicit, resolves the issue: https://koji.fedoraproject.org/koji/taskinfo?taskID=118918826

PR #553, opened 2024-06-12
zuul: Update versions of ansible and ansible-core tested

dmsimard:
We're up to Ansible 10 and ansible-core 2.17.

### DIFF
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -113,13 +113,13 @@
         override-checkout: devel
 
 - job:
-    name: ara-basic-ansible-9
+    name: ara-basic-ansible-10
     parent: ara-ansible-integration-base
     description: |
-      Runs basic integration tests through the equivalent of "tox -e ansible-integration" with ansible 8 (ansible-core 2.15).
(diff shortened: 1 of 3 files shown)


dmsimard:
There is a change of behavior in ansible-core 2.17 where previously-expected vars files were saved and now they aren't:

## 2.16.7
Screenshot from 2024-06-14 21-28-01

## 2.17.0
Screenshot from 2024-06-14 21-27-56

dmsimard:
Now, it's not like ara relies on a stable ansible-core API to retrieve those files:
https://github.com/ansible-community/ara/blob/a7fc0a8f8a74add07b780e7e44276589334827d2/ara/plugins/callback/ara_default.py#L502-L514

There might be a better way to do this now, I will need to investigate.

dmsimard:
I spent some time trying to understand the change in behavior and created an issue about it: https://github.com/ansible-community/ara/issues/560

I'm not sure what to do with it for now but it's not an issue specific to the upcoming release of ara 1.7.0.
Any version of ara running with ansible-core 2.17 may notice a change in the number of saved files but the playbook and task files should still remain recorded for sure.

PR #557, opened 2024-06-22
docs: update requirements to include ara

dmsimard:
The sphinx docs were broken on readthedocs, seemingly due to an expectation that ara was installed by default.

The docs includes output from every ara --help command and so sphinx must have ara installed for that to work.

This has always been true and I don't remember why the requirements were split up in this way.


dmsimard:
## Before
Screenshot from 2024-06-21 21-00-11

## After
Screenshot from 2024-06-21 20-59-11

PR #558, opened 2024-06-22
Docs refresh

dmsimard:
The getting started guide was originally a copy from the README
and it accidently became stale and out of date as the README was
improved and maintained.

We can make the README even better eventually and there shouldn't
be two sources of truth.

Let's keep the README for this and link to it instead.

ara-web has been archived in February 2023, update the docs to remove
it and add ara-collection which was missing.

Take the opportunity to refresh the question about different components
of ara running on different servers.

### DIFF
--- a/doc/source/getting-started.rst
+++ b/doc/source/getting-started.rst
@@ -1,125 +1,20 @@
 Getting started
 ===============
 
-Requirements
-------------
+The `README <https://github.com/ansible-community/ara/blob/master/README.md>`_ is the best place to get started.
 
-- Any recent Linux distribution or Mac OS with python >=3.5 available
-- The ara Ansible plugins must be installed for the same python interpreter as Ansible itself
+It will tell you about:
 
-.. note::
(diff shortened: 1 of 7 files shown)


doc/source/getting-started.rst
```diff
@@ -1,125 +1,20 @@
...
-ara is fully packaged for Fedora_, OpenSUSE_ as well as Debian_.
-
-There is also a package without the API server available on RHEL 8/CentOS 8 EPEL_.
-This package contains the Ansible plugins as well as the API clients which are sufficient to query or send data to a
-remote API server.
-
-.. _Fedora: https://koji.fedoraproject.org/koji/packageinfo?packageID=24394
-.. _OpenSUSE: https://build.opensuse.org/package/show/devel:languages:python/python-ara
-.. _Debian: https://tracker.debian.org/pkg/python-ara
-.. _EPEL: https://koji.fedoraproject.org/koji/packageinfo?packageID=24394
```
dmsimard, line 117: This is worth re-mentioning elsewhere.

Issue #556, opened 2024-06-20
Catch SIGTERM and SIGINT


Nudin:
## What is the idea ?

When an ansible deploy is stopped – by for example C^c, kill or a crash of the controller – ara will continue to show the run as "running" indefinitely or until `ara expire` is executed. For some kinds of crashes like SIGKILL or kernel panics this is unavoidable, but for the most common cases SIGINT (C^c) and SIGTERM it would be possible to catch the signal to gracefully mark the run as failed.  This would improve the usefulness of ara, since one would see if a run is still ongoing or was canceled by the caller.

I tried adding a signal trap to the ara ansible callback script and it works.

dmsimard:
Hi @Nudin and thanks for the issue.

This use case is in part why the [ara expire](https://ara.readthedocs.io/en/latest/cli.html#ara-expire) command exists: to eventually expire playbooks that crashed or were interrupted and thus remained in running forever.

I like the idea of trying to capture signals if it is possible and you have something working.

I wonder how reliable it would be because running from within the ansible callback plugin interface is kind of special but it is probably better than nothing.

I also wonder about whether the status of the playbook should be failed, expired or if we need something new like "interrupted" because the playbook didn't /really/ fail.

If you would like to open a PR with what you have we can discuss it and go from there.

bendem:
This is done, it can probably be closed.

PR #559, opened 2024-06-27
Catch SIGINT and SIGTERM (#556)

Nudin:
This will catch SIGINT/SIGTERM and send an update to the ara server, so
that the playbook does not continue to be shown as running indefinitely
but will report failed. See #556

After sending the update to the server, it executes any handler that where
configured so far in order not to interfere with other parts of
ansible/ansible plugins or if none was present, exit with the appropriate exit code.

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -8,6 +8,7 @@ import getpass
 import json
 import logging
 import os
+import signal
 import socket
 import sys
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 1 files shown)


dmsimard:
Hi and thanks for the PR!

I am interested in testing it and will circle back to it sometime soon to make sure it's included in the next release of ara.

ara/plugins/callback/ara_default.py
```diff
@@ -327,6 +328,27 @@ def __init__(self):
...
 
+        self.set_handler(signal.SIGINT)
+        self.set_handler(signal.SIGTERM)
+
+    def set_handler(self, sig):
+        original_handler = signal.getsignal(sig)
+
+        def handler(sig, frame):
+            ended = datetime.datetime.now(datetime.timezone.utc).isoformat()
+            status = "failed"
```
dmsimard, line 339: I tested this and it works, thanks !

I hesitate on whether failed is the appropriate status to use.

The status will be failed but there won't really be a failed task to look at which may be confusing without additional information.

I see two options:
- We re-use the existing ``expired`` status which, like mentioned in the issue, comes from expiring those interrupted or unfinished playbooks that will never finish
- We create a new status, ``interrupted`` to make it clear that this particular playbook was interrupted

Creating a status is not particularly difficult but it is not free either.
We need to adjust the database model, the API, the CLI and the UI accordingly.

Re-using ``expired`` is essentially free and I feel it's a better representation of what happened: the playbook didn't fail but it will also never finish.

This is what it'd look like using expired vs failed:
Screenshot from 2024-06-30 17-35-11

What do you think ?

dmsimard: I'd like to ship this in a release soon. I will avoid creating a new status for now and use the existing expired one.

dmsimard:
I have submitted a PR with an update to use the expired status instead: https://github.com/ansible-community/ara/pull/565

I will close this one.

Thanks for the nice improvement and sorry it took so long to get around to it.

PR #567, opened 2024-08-26
client: allow using utf-8 usernames and passwords

bendem:
It looks like requests supports utf-8, but only if it is pre-encoded by the caller. Otherwise, it tries to encode both parameters using latin-1.

See 
https://github.com/psf/requests/blob/a3ce6f007597f14029e6b6f54676c34196aa050e/src/requests/auth.py#L56-L60 https://github.com/psf/requests/issues/3662
https://github.com/pallets/werkzeug/issues/945

### DIFF
--- a/ara/clients/utils.py
+++ b/ara/clients/utils.py
@@ -22,7 +22,7 @@ def get_client(
     """
     auth = None
     if username is not None and password is not None:
-        auth = HTTPBasicAuth(username, password)
+        auth = HTTPBasicAuth(username.encode('utf-8'), password.encode('utf-8'))
 
     # Verify can be a bool (to ignore SSL verification or not)
     # or the path to a certificate authority
--- a/ara/clients/utils.py
+++ b/ara/clients/utils.py
@@ -22,7 +22,7 @@ def get_client(
     """
     auth = None
     if username is not None and password is not None:
-        auth = HTTPBasicAuth(username.encode('utf-8'), password.encode('utf-8'))
+        auth = HTTPBasicAuth(username.encode("utf-8"), password.encode("utf-8"))
 
     # Verify can be a bool (to ignore SSL verification or not)
     # or the path to a certificate authority


ara/clients/utils.py
```diff
@@ -22,7 +22,7 @@ def get_client(
     """
     auth = None
     if username is not None and password is not None:
-        auth = HTTPBasicAuth(username, password)
+        auth = HTTPBasicAuth(username.encode('utf-8'), password.encode('utf-8'))
```
ipoac2549, line 25:
```suggestion
        auth = HTTPBasicAuth(username.encode("utf-8"), password.encode("utf-8"))
```
The linter is complaining that these should be double quotes.

dmsimard, line 25:
```suggestion
        auth = HTTPBasicAuth(username.encode("utf-8"), password.encode("utf-8"))
```
The linter is complaining because it wants double quotes.

dmsimard:
Hi and thanks for the PR.

I'd like to reproduce the issue and test that the fix works since we don't have an integration test for something like this yet.

Could you give an example of username/password that we could test with ?

bendem:
Sure thing, I've managed to reproduce this by running a dummy playbook with both user and password env vars set to €. No server needed

27 Aug 2024 06:02:34 David Moreau Simard ***@***.***>:

dmsimard:
Reproducing the issue without the patch:
```
PLAYBOOK: hosts.yaml **************************************************************************************************************************************************************************
[WARNING]: Failure using method (v2_playbook_on_start) in callback plugin (<ansible.plugins.callback.ara_default.CallbackModule object at 0x7f73b6487a70>): 'latin-1' codec can't encode
character '\u20ac' in position 0: ordinal not in range(256)
Callback Exception:
  File "ara/.tox/ansible-integration/lib/python3.12/site-packages/ansible/executor/task_queue_manager.py", line 450, in send_callback
    method(*new_args, **kwargs)
   File "ara/ara/plugins/callback/ara_default.py", line 468, in v2_playbook_on_start
    self.playbook = self.client.post(
                    ^^^^^^^^^^^^^^^^^
   File "ara/ara/clients/http.py", line 105, in post
    return self._request("post", endpoint, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "ara/ara/clients/http.py", line 83, in _request
    response = func(url, **kwargs)
               ^^^^^^^^^^^^^^^^^^^
   File "ara/ara/clients/http.py", line 56, in post
    return self._request("post", url, data=json.dumps(payload))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "ara/ara/clients/http.py", line 44, in _request
    return self.http.request(method, self.endpoint + url, timeout=self.timeout, **payload)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "ara/.tox/ansible-integration/lib/python3.12/site-packages/requests/sessions.py", line 575, in request
    prep = self.prepare_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
   File "ara/.tox/ansible-integration/lib/python3.12/site-packages/requests/sessions.py", line 486, in prepare_request
    p.prepare(
   File "ara/.tox/ansible-integration/lib/python3.12/site-packages/requests/models.py", line 372, in prepare
    self.prepare_auth(auth, url)
   File "ara/.tox/ansible-integration/lib/python3.12/site-packages/requests/models.py", line 603, in prepare_auth
    r = auth(self)
        ^^^^^^^^^^
   File "ara/.tox/ansible-integration/lib/python3.12/site-packages/requests/auth.py", line 95, in __call__
    r.headers["Authorization"] = _basic_auth_str(self.username, self.password)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "ara/.tox/ansible-integration/lib/python3.12/site-packages/requests/auth.py", line 57, in _basic_auth_str
    username = username.encode("latin1")
               ^^^^^^^^^^^^^^^^^^^^^^^^^
2 plays in hosts.yaml
```

I must point out that django doesn't seem to like it:
```
> ara-manage createsuperuser --username=€uro --email=joe@example.com
[ara] Using settings file: /home/dmsimard/.ara/server/settings.yaml
Enter a valid username. This value may contain only letters, numbers, and @/./+/-/_ characters.
```

htpasswd and nginx do not seem to mind:
```
> sudo htpasswd -c -m /etc/nginx/.htpasswd €uro
New password: 
Re-type new password: 
Adding password for user €uro

> cat /etc/nginx/.htpasswd 
€uro:$apr1$yxshm3eD$ulGhdUVTRuWeAHGpUhEAD0
```

```
127.0.0.1 - \xE2\x82\xACuro [28/Aug/2024:21:32:06 -0400] "GET / HTTP/1.1" 502 157 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:129.0) Gecko/20100101 Firefox/129.0" "-"
```

I get no error with the patch included, thanks @bendem :)

Could you update your PR or merge my suggestion to satisfy the linter ?
We can merge it afterwards.

bendem:
Sorry, I've been busy for a few days, I committed your suggestion. Should be all good.



> I must point out that django doesn't seem to like it:

We use external auth, so I didn't see this. I understand they would want to restrict user names, but in our case, the problem was with the password which really shouldn't be restricted.

dmsimard:
It's merged, thanks for the patch :)

1.7.2 was supposed to be released this week but I will spin up a new release candidate to include the fix right away.

Issue #283, opened 2021-06-30
ARA authentication slows down ansible


pescobar:
I have deployed  ARA and I have noticed that the ansible execution time increases a lot when I enable ARA authentication. 

The deployment is is based on docker, using the official ARA image, postgres as db backend and nginx as a reverse proxy in front. I am using this compose file slightly modified https://gist.github.com/pescobar/c4668bfc4f8bec86bdab06cb3eae3428

The ARA server is located in a different machine so the ansible controller is pushing the data over the network using https.

All the tests are done executing the same playbook. 

This is the ansible execution time without ARA:
```
real    6m29.206s
user    2m30.816s
sys     0m44.413s
```

This is the ansible execution time using ara WITHOUT authentication:
```
real    7m34.077s
user    2m38.239s
sys     0m48.011s
```

This is the ansible execution time using ara WITH authenticaion:
```
real    13m6.823s
user    2m38.622s
sys     0m50.371s
```

Is there any workaround for this? Any setting I can tweak to improve performance when ara authentication is enabled?

dmsimard:
Hi @pescobar and thanks for the issue.

I'll preface this with a blog post that I have written a while back to calculate the performance overhead of the callback: https://ara.recordsansible.org/blog/2020/11/01/benchmarking-ansible-and-ara-for-fun-and-science/

It turns out that I had not benchmarked with and without authentication so thanks for providing some metrics.

I am not well informed about the potential overhead of supplying authentication via python's requests library and how it might impact django performance but we can learn together :P

First, some questions:
- When using the http API client and the postgre database backend, you can enable callback threading to improve performance with [ARA_CALLBACK_THREADS](https://ara.readthedocs.io/en/latest/ansible-plugins-and-use-cases.html#ara-default-callback-recording-playbooks-from-ansible). Do you have this enabled and does it help ?
- What are you using as the authentication backend ? Is it regular django with superusers and users or are you doing something with nginx ? I ask because some users leverage [ARA_EXTERNAL_AUTH](https://ara.readthedocs.io/en/latest/api-configuration.html#ara-external-auth) and then manage authentication via apache instead (i.e, mod_ldap).

pescobar:
Thanks for your reply @dmsimard, I had missed the option ` ARA_CALLBACK_THREADS`

I am using django as authentication backend. I added the users to django as described in the docs.

I did some tests increasing the number of threads and these are the numbers:

without authentication and 4 threads
```
real    6m56.702s
[...]

PR #570, opened 2024-09-19
docs: clarify troubleshooting performance with Django authentication

Frazew:
# Summary

This PR is related to  #283. Feel free (of course) to close if you think it's not necessary!

Using the builtin Django authentication system on the API server results in very significant performance degradation. The documentation already strongy suggests delegating authentication to a load balancer / reverse proxy but some people might still rely on the default authentication while experimenting.

The Troubleshooting section hinted that the performance issue could be due to Django<->database latency, when it is in fact due to the use of PBKDF2 for password hashing.


doc/source/troubleshooting.rst
```diff
@@ -131,7 +131,7 @@ In no particular order, here's high-level advice that have proven to be useful i
 - The built-in Django development server provided by the default offline API client and ``ara-manage runserver`` is simple and convenient but it isn't meant to provide the best scalability and performance
 - The API server should be run as a service with a WSGI application server like gunicorn, uwsgi or mod_wsgi with apache2/httpd
 - There should be a frontend, reverse-proxy or load balancer such as apache, nginx, haproxy or traefik in front of the API server in order to handle TLS termination, caching and authentication
-- When :ref:`enabling authentication <api-security:Authentication and user management>`, consider using ``EXTERNAL_AUTH`` instead of the Django built-in user management to avoid a database lookup performance hit on every query
+- When :ref:`enabling authentication <api-security:Authentication and user management>`, consider using ``EXTERNAL_AUTH`` instead of the Django built-in user management to avoid the very significant performance hit of password hashing on every request
```
dmsimard, line 134: Thanks for the clarification.

This isn't inaccurate but my understanding is that there is still a database lookup involved (it's where the password is stored). It's also OK to link to the issue for more context, thus I would suggest something like:

```suggestion
- When :ref:`enabling authentication <api-security:Authentication and user management>`, consider using ``EXTERNAL_AUTH`` instead of the Django built-in user management. This will avoid a significant performance hit incurred while hashing the password when authentication must be validated against the database backend. (See issue `#283 <https://github.com/ansible-community/ara/issues/283>`_)
```

Frazew: you're absolutely right, I didn't realize I completely dropped the hint about the database query, thank!

dmsimard:
LGTM, thanks !

PR #574, opened 2024-11-04
container-images: bump fedora images to fedora:40

dmsimard:
39 is starting to be old with 41 just recently released.

We will eventually update to 41, let's just bump a single release for now.

### DIFF
--- a/contrib/container-images/fedora-source.sh
+++ b/contrib/container-images/fedora-source.sh
@@ -3,7 +3,7 @@
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 DEV_DEPENDENCIES="gcc python3-devel postgresql-devel mariadb-devel"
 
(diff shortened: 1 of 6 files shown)


contrib/container-images/fedora-source.sh
```diff
@@ -3,9 +3,9 @@
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 DEV_DEPENDENCIES="gcc python3-devel postgresql-devel mariadb-devel"
 
-# Builds an ARA API server container image from checked out source on Fedora 39.
+# Builds an ARA API server container image from checked out source on Fedora 40.
 # Figure out source directory relative to the contrib/container-images directory
-SCRIPT_DIR=$(cd `dirname $0` && pwd -P)
+SCRIPT_DIR=$(cd $(dirname $0) && pwd -P)
```
dmsimard, line 8: Unintended change by linter, it should remain that way.

PR #579, opened 2025-01-10
Client: Enable to ignore all facts

superstes:
The Ansible-facts `gather_subset` allows for the value `all` to be supplied: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/setup_module.html#parameter-gather_subset

It would be logical that ARA allows users to ignore all facts the same way.
This PR adds that functionality.

This is especially useful as some facts are dynamically named - like network-interface facts.

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -848,7 +848,7 @@ class CallbackModule(CallbackBase):
                 for fact in self.ignored_facts:
                     if fact in results["ansible_facts"]:
                         self.log.debug("Ignoring fact: %s" % fact)
-                        results["ansible_facts"][fact] = "Not saved by ARA as configured by 'ignored_facts'"
+                        results["ansible_facts"][fact] = ignored_facts_hint
 
         # Note: ignore_errors might be None instead of a boolean
         ignore_errors = kwargs.get("ignore_errors", False) or False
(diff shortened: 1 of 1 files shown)


dmsimard:
Hi, thanks for the PR.

This seems reasonable. I'll test it and consider merging it before the next release.

ara/plugins/callback/ara_default.py
```diff
@@ -838,10 +838,17 @@ def _load_result(self, result, status, **kwargs):
...
         if "ansible_facts" in results:
-            for fact in self.ignored_facts:
-                if fact in results["ansible_facts"]:
-                    self.log.debug("Ignoring fact: %s" % fact)
-                    results["ansible_facts"][fact] = "Not saved by ARA as configured by 'ignored_facts'"
+            ignored_facts_hint = "Not saved by ARA as configured by 'ignored_facts'"
+
+            if "all" in self.ignored_facts:
+                self.log.debug("Ignoring all facts")
+                results["ansible_facts"] = {"all": ignored_facts_hint}
```
dmsimard, line 845: I kind of like that we don't iterate across every fact to set the values individually. Good thinking.

dmsimard:
Works for me, thanks !

### Before
Screenshot From 2025-02-08 14-24-57

### After, with ``export ARA_IGNORED_FACTS=all``
Screenshot From 2025-02-08 14-25-01


I will create a follow-up PR to add a bit of docs and see how we might do an integration test to make sure this works.

Issue #581, opened 2025-01-20
ara_playbook_labels not added


NiceRath:
## What is the issue ?
We try to pass some inventory-specific & group-specific labels to the ARA-client.

To ensure these are set as expected - we assert it inside our playbook:

```yaml
- name: Test
  hosts: all
  become: true
  gather_facts: true
  pre_tasks:
    - ansible.builtin.assert:
        that:
          - ara_playbook_labels is defined
          - ara_playbook_labels | length > 0
```

But when I check-out the `play_vars` - the variable is not set?
```python3
...
raise ValueError("ara_playbook_labels" in play_vars, play_vars.keys())
if "ara_playbook_labels" in play_vars:
    ...
```
At: https://github.com/ansible-community/ara/blob/master/ara/plugins/callback/ara_default.py#L511

It Prints:

```
(False, dict_keys([<USER-SPECIFIC-VARS-REDACTED>, 'playbook_dir', 'ansible_playbook_python',
'ansible_config_file', 'ansible_role_names', 'ansible_play_role_names', 'ansible_dependent_role_names', 'role_names',
'ansible_play_name', 'groups', 'ansible_play_hosts_all', 'ansible_play_hosts', 'ansible_play_batch', 'play_hosts', 'omit',
'ansible_version', 'ansible_check_mode', 'ansible_diff_mode', 'ansible_forks', 'ansible_inventory_sources',
'ansible_skip_tags', 'ansible_limit', 'ansible_run_tags', 'ansible_verbosity', 'hostvars']))
```

When in the play_vars - the variable exists. But it seems to be nested under `hostvars` .

```python3
...
for k, v in play_vars.items():
    if v is not None and str(v).find("ara_playbook_labels") != -1:
        raise ValueError(k, v)
...
```

## What should be happening ?

I would expect the variable to work as documented: https://ara.readthedocs.io/en/latest/ansible-plugins-and-use-cases.html#playbook-names-and-labels

Two solutions come to my mind:
* Either update the documentation, so it mentions that inventory-vars DO NOT work.
* Or check the hostvars for `ara_playbook_labels`

NiceRath:
I understand that these labels are not host-specific, but specific to the execution.
But we should be able to iterate through the `hostvars` and extend the list of labels by all that we find..

superstes:
PR: https://github.com/ansible-community/ara/pull/582

dmsimard:
Hi and thanks for the issue.

[...]

PR #582, opened 2025-01-22
Client: Enable to supply labels as inventory-vars

superstes:
ARA-Labels can currently only be set inside a playbook or as extra-vars.

As Ansible uses the inventory-system to dynamically set variables - it would make sense to allow users to set those labels inside their inventory.
This PR adds exactly that functionality.

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -518,6 +518,16 @@ class CallbackModule(CallbackBase):
                 labels.extend(play_vars["ara_playbook_labels"].split(","))
             else:
                 raise TypeError("ara_playbook_labels must be a list or a comma-separated string")
+
+        for host in play_vars["ansible_play_hosts"]:
+            if "ara_playbook_labels" in play_vars["hostvars"][host]:
+                if isinstance(play_vars["hostvars"][host]["ara_playbook_labels"], list):
+                    labels.extend(play_vars["hostvars"][host]["ara_playbook_labels"])
+                elif isinstance(play_vars["hostvars"][host]["ara_playbook_labels"], str):
+                    labels.extend(play_vars["hostvars"][host]["ara_playbook_labels"].split(","))
+                else:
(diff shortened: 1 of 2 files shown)


superstes:
It was unclear to me how integration tests should be added.
Might make sense to add a test-cases for it.

superstes:
The docs-test error/warning seems to be unrelated (?):
`WARNING: Calling get_html_theme_path is deprecated. If you are calling it to define html_theme_path, you are safe to remove that code.`

superstes:
Note: We could of course deduplicate the `isinstance` and `split` checks.

dmsimard:
Hi, I added a comment relevant to this PR in the issue: https://github.com/ansible-community/ara/issues/581#issuecomment-2614666763

I'd like that we consider what is the right solution because iterating over every host is expensive.

dmsimard:
I will leave this PR open for now but I would rather not go with the suggested implementation for now.

When I have more time I would like to benchmark this approach to demonstrate my point.

superstes:
If performance of the for-loop is the possible stopper:
Might we be able to use a config-option to opt-in to this functionality?

The labels system might be a lot more useful if inventory-vars could be used 🤔

dmsimard:
> If performance of the for-loop is the possible stopper: Might we be able to use a config-option to opt-in to this functionality?
> 
> The labels system might be a lot more useful if inventory-vars could be used 🤔

This came back on my radar recently and may be of interest to you: https://github.com/ansible-community/ara/issues/596

dmsimard:
Hi,

The ara_label module will ship in ara 1.7.3 soon which may provide a workaround to set up labels from vars.
I think you could loop through host vars and use ara_label or something like that.

In the meantime the project is migrating to Codeberg: https://github.com/ansible-community/ara/issues/603
This pull request was migrated to Codeberg: https://codeberg.org/ansible-community/ara/pulls/582

I am as such closing this PR.

Feel free to comment here or on Codeberg, this repository is not going away and will instead become a mirror.
Merging of pull requests will happen on Codeberg, though.

PR #585, opened 2025-01-26
Client: Add ignored_results option

jadyndev:
This adds the option to ignore any results returned by an ansible task based on user supplied regexes.
This way you can customize which data passed from and to ansible tasks gets forwarded to ara.

The variables are filtered based on their path: Which consists of the tasks action and the dot-notated path to the variable inside the `results` dict: `ansible.builtin.user:invocation.password` 

Adding `ignored_results = .*password.*` would be enough to filter all variables containing `password` in their path, due to the flexibility given by regex, everything from just filtering specific variables, modules/plugins to whole collections is now possible.

### DIFF
--- a/ara/plugins/callback/ara_default.py
+++ b/ara/plugins/callback/ara_default.py
@@ -8,9 +8,11 @@ import getpass
 import json
 import logging
 import os
+import re
 import signal
 import socket
 import sys
+from collections.abc import MutableMapping
 from concurrent.futures import ThreadPoolExecutor
 
 from ansible import __version__ as ANSIBLE_VERSION, constants as C
@@ -190,6 +192,21 @@ options:
     ini:
       - section: ara
         key: ignored_files
+  ignored_results:
+    description:
+      - List of regexes which are matched against variable paths
+      - Paths consist of the task_action (as defined in your playbook) and the variables used by the task.
+      - "Example: '<task_action>:<dict_path>' -> 'ansible.builtin.user:invocation.password'"
+      - This regex 'ansible\.builtin\.user:invocation\.password' would remove the password from the result
+        and replace it with a message
+      - Ensure that each '.' in the variable path is escaped with '\.'
+    type: list
+    default: []
+    env:
+      - name: ARA_IGNORED_RESULTS
+    ini:
+      - section: ara
+        key: ignored_results
   localhost_as_hostname:
     description:
         - Associates results to the hostname (or fqdn) instead of localhost when the inventory name is localhost
@@ -312,6 +329,7 @@ class CallbackModule(CallbackBase):
         self.ignored_facts = []
         self.ignored_arguments = []
         self.ignored_files = []
+        self.ignored_results = []
         self.localhost_as_hostname = None
         self.localhost_as_hostname_format = None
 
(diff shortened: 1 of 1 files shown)


dmsimard:
Hi and thanks for the PR.

I need to think a little bit about this one.

There's a big regex that hides somewhere in Ansible that tries to have an heuristic like this to try and prevent tokens, passwords and such from being displayed.

I'll look into it. Thanks.

jadyndev:
Hi,
I can give you a bit of context on why I made this PR:
My ansible not only configures things via ssh, I make a bunch of API calls: to netbox, in case data needs updating; to my DNS-Servers; etc. 
As well as API calls I also generate/load a bunch of secrets on-demand, which happens outside of ansible-vault.

Therefore many secrets, passwords,  api-keys, etc. are passed from and to tasks, sometimes also in ansible_facts.
I've been introduced to ara early last year, but as it would log pretty much all my secrets I was very uncomfortable to use it (outside my PC). I want to use ara to monitor/debug auto-deployments, which would force me to host ara somewhere. And as I usually take care only to store secrets encrypted (with keys on smartcards) ara would punch a deep hole in my security. This feature would solve this issue as I can easily stop ara from reporting data on certain tasks handling secrets.

dmsimard:
Yeah, that makes sense. Thanks for the context.

Someone helped me find that heuristic in Ansible again: https://github.com/ansible/ansible/blob/e6adddcaf8d4cf46185a90be89a574ba01cc6b84/lib/ansible/module_utils/basic.py#L164

That regex tries to prevent sensitive things from being displayed in the console but it's... imperfect. It will never catch 100% of the sensitive things. I feel like taking a similar approach in ara would be similarly imperfect. I wouldn't let that be the enemy of "good enough" but I am putting that out there as part of the thought process.

I mean to say that ara records ansible, so if the data makes it in to ara, it could perhaps leak in your console/terminal/job as well. Making it so ara won't record them is only half the story, and making sure that Ansible doesn't send that data to ara at all in the first place would be best.

There is ``no_log: true`` set on tasks for this, but it can be inconvenient when the time comes to troubleshoot or understand what the task did. It is the one 100% sure way that a particular task wouldn't be recorded, though.

I am not declining this PR for now, I just need to think a little bit more about this, and ponder on what a feature like this might look like in upstream Ansible such that the sensitive data doesn't make it into ara in the first place.

flowerysong:
Data tagging is coming (any year now) and one of the things it will do (in theory, there's no public implementation available yet) is provide a more comprehensive approach to dealing with sensitive values.

dmsimard:
> Data tagging is coming (any year now) and one of the things it will do (in theory, there's no public implementation available yet) is provide a more comprehensive approach to dealing with sensitive values.

Yes, this appears to be the takeaway from briefly discussing it in the Ansible Social channel today.

dmsimard:
@flowerysong I won't link to it directly here (because I see no need to have github set up a link with this issue) but there is a new PR about data tagging: https://github.com/ansible/ansible/pulls?q=is%3Apr+is%3Aopen+%5BWIP%5D+Templating+overhaul%2C+implement+Data+Tagging

dmsimard:
It seems like data tagging is being prioritized to the extent that core is holding back on merging PRs to prevent rebase issues.

Given that the results of that are unknown to me at this time I will leave this PR open for now to make sure we choose the right way forward. Thanks for your patience.

dmsimard:
By the way I've created an issue more specifically about data tagging: https://github.com/ansible-community/ara/issues/588

dmsimard:
I'm not sure datatagging brings anything to the table to help with stuff like this and I am still not convinced in trying to filter things with regexes but I am open to suggestions.

In the meantime, the project is migrating to Codeberg: https://github.com/ansible-community/ara/issues/603
This PR has been migrated there: https://codeberg.org/ansible-community/ara/pulls/585

I am as such closing this PR.

Feel free to comment here or on Codeberg, this repository is not going away and will instead become a mirror.
Merging of pull requests will happen on Codeberg, though.

Issue #295, opened 2021-07-24
UI: Add an auto-refresh to automatically refresh the page after a given duration


dmsimard:
## What is the idea ?

The built-in reporting UI doesn't have fancy asynchronous javascript that automatically pulls and update data without refreshing the page.

As a stopgap, it would be a nice alternative to allow for the page to refresh automatically every N seconds or N minutes.

It could look like this:
![Peek 2021-07-23 23-32](https://user-images.githubusercontent.com/1291204/126856227-33d86925-5e56-4792-934a-cdc9cf7d52fc.gif)

This could make the interface easier to use as a "live dashboard" that is always up to date.

gaby:
Any updates on adding this feature? I'm constantly refresh to see the status of playbooks running. Even worst when running against multiple hosts

dmsimard:
I agree that this would be nice to have :+1:

Looking at the mockup I did last year, I think I would do it differently today.

I think my idea at the time was to set a variable in the URL (say, ``autorefresh=60``) for a value in seconds via those links in the dropdown menu.

It's probably way more convenient to set the refresh value in a cookie so it follows the user across pages easily without needing to include it in the URL everywhere. We don't set any cookies right now but there's ways to do it with django: https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpResponse.set_cookie

Anyway, on refresh, we'd add a meta refresh with the value of the variable like this:
```
<meta http-equiv="refresh" content="{{ autorefresh }}">
```

Such that the browser refreshes the page automatically -- no need for javascript or anything else.

I don't have the code for the dropdown anymore and I can't find the branch for it but there's bootstrap docs about them here: https://getbootstrap.com/docs/4.6/components/dropdowns/

A drawback of that implementation is that it requires a new javascript dependency (popper.js) and it's harder to see whether auto-refresh would be enabled and what value it is set at.

I've tried thinking how we should implement it and I've come up with two mockups:
![Peek 2022-08-11 22-27](https://user-images.githubusercontent.com/1291204/184274605-be2c4386-fef2-4450-9caa-855f95dc7706.gif)

![Peek 2022-08-11 22-35](https://user-images.githubusercontent.com/1291204/184274643-bcf5e3ef-ef18-43b7-91db-3377c7d007ff.gif)

I thought the slider would be nice but it's harder to be precise in value.

What do you think ? What should it look like ?

[...]

PR #590, opened 2025-02-08
ui: Add an auto-refresh feature

dmsimard:
An arrow drop-down menu at the top right that
auto refreshes ara (another recursive acronym).

- Disabled by default
- Duration selection saved across loads in local storage
- 3 seconds before a refresh, the arrow turns green and spins

### DIFF
--- a/ara/ui/static/css/ara.css
+++ b/ara/ui/static/css/ara.css
@@ -115,3 +115,23 @@
     width: 25px;
     text-align: center;
 }
+
+/*
+ * Auto-refresh styling
+*/
+/* Spin and color the refresh arrow */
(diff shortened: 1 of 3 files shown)


dmsimard:
This is one of the oldest issues in the GitHub repo, dating back to 2021: https://github.com/ansible-community/ara/issues/295

Here's what it looks like:
Screenshot From 2025-02-07 23-53-38
Screenshot From 2025-02-07 23-53-33

dmsimard:
Also tested that it works with the static generation, though there isn't much of a point to auto-refreshing a static build.. unless the pages could change underneath ?

image

PR #591, opened 2025-02-08
ignored_facts: Add a bit of docs and tests

dmsimard:
Following the addition of supporting "all" as the value of ignored_facts, we should take the opportunity to document the feature and then test that it works.


dmsimard:
@superstes: as a follow up to https://github.com/ansible-community/ara/pull/579

PR #592, opened 2025-02-08
CI: bump fedora, ansible and ansible-core versions

dmsimard:


dmsimard:
It seems like there's a behavior change in ansible-core 2.18 where it previously allowed non-utf8 characters to be returned but now there's an exception:
image

Issue #598, opened 2025-03-31
ara cli doesn't use ARA_BASE_PATH in API calls


betrotle:
## What is the issue ?

The ara cli fails to connect to an ara API instance where a BASE_PATH configured. E.g.

root@1c62864d979c:/opt/ara# export ARA_API_CLIENT=http
root@1c62864d979c:/opt/ara# export ARA_API_SERVER=http://ara-api-1:8000
root@1c62864d979c:/opt/ara# export ARA_BASE_PATH=/ara/
root@1c62864d979c:/opt/ara# ara playbook list
Failed to get on /api/v1/playbooks: {'params': {'order': '-started', 'limit': 50}}
Expecting value: line 1 column 1 (char 0)

## What should be happening ?
ara cli should update the endpoint being queried to include the BASE_PATH

flowerysong:
This does not appear to be a bug; `ARA_BASE_PATH` is not documented as being an environment variable accepted by the client.

While I don't have an environment where I can easily test this, I would expect to specify the API location is as part of the URL, not a separate variable: `export ARA_API_SERVER=http://ara-api-1:8000/ara`

betrotle:
Hi flowerysong, confirmed appending the API location as part of the URL does work, thanks for your quick response!

In any case, there's a pull request https://github.com/ansible-community/ara/pull/599 if wanting to support ARA_BASE_PATH which better aligns with the ARA API server config (https://ara.readthedocs.io/en/latest/api-configuration.html#ara-base-path)

PR #599, opened 2025-03-31
ara cli to use ARA_BASE_PATH

betrotle:
To resolve issue raised in https://github.com/ansible-community/ara/issues/598


dmsimard:
Hi and thanks for the PR.

Like mentioned in the linked issue, ARA_BASE_PATH is a django server environment variable.

When setting up ARA_API_SERVER for the client/callback/cli it should include the full URL including the base path if you decided to use one.

I understand why it might have been confusing, we can probably improve the documentation but we aren't looking to include this variable for the client at this time.

